PR #5, opened 2015-11-03
GPII-1336: update `schemaMiddleware` to work with more than just POST data...

the-t-in-rtf:
See [GPII-1336](https://issues.gpii.net/browse/GPII-1336) for details.


.jshintrc
```diff
@@ -0,0 +1,32 @@
...
+    "unused":    true,
+    "strict":    true,
+    "trailing":  true,
+    "white":     true,
+
+    "maxerr": 1000,
+
+    "node": true,
+
+    "globals": {
```
amb26: This block is only used in client-side code and can be removed

README.md
```diff
@@ -1 +1,133 @@
...
+JSON Schemas provide a clear way to indicate what is and is not valid within a given JSON document.  You might use JSON
+Schemas to validate a form's data on the client side before attempting to submit it to the server.  You might also
+validate the data whenever the model changes.
+
+This package provides client and server side validators, which share a common base package.  To use this on the client
+side, start with `gpii.schema.validator.client`.  To use it on the server side, start with `gpii.schema.validator.server`.
+Each package contains more detailed documentation about its use.  Here is an example of how it might work from within
+node:
+
+  var fluid = fluid || require("infusion);
```
amb26: Multiple typos here - stray quote and client-side require

README.md
```diff
@@ -1 +1,133 @@
...
+other schemas.  The validator in this package supports references between JSON Schemas.
+
+Inheritance is still a sticking point at least in v4 of the draft standard.   You cannot safely expect to overlay
+multiple schemas on top of each other.  Best practice for now is to only reuse individual definitions between schemas,
+and to explicitly specify each schema's required properties.
+
+A limitation of this implementation is that it expects external references to use filenames rather than ids,
+as in:
+
+    $ref: "filename.json#/definitions/field`
```
amb26: Stray quote here

README.md
```diff
@@ -1 +1,133 @@
...
+
+This is a by product of the library we use to evolve the error output. As this may change, best practice is for the
```
amb26: byproduct one word - best to name the library here too

README.md
```diff
@@ -1 +1,133 @@
...
+
+Although you usually will build some fault tolerance into your components, on some level they expect to deal with
+data that has the right structure, and which contains the expected type of information (strings, booleans, dates, etc).
+
+The `gpii.schema.middleware` component provided with this package rejects invalid payloads, which allows your server-side
+components to safely assume they will only receive JSON data in the correct format.
+
+The component is intended to be wired into an existing rest endpoint as in the following example:
+
+    var fluid = fluid || require("infusion");
```
amb26: include

README.md
```diff
@@ -1 +1,133 @@
...
+(`gpii.schema.validator.server`), and is intended to be used in conjunction with a `gpii.express.handler`.  For
+examples, see the tests in this package.
+
+# Using the validator in a browser
+
+To use this component in a browser, you will need to run `browserify` against ajv and generate a client-side bundle,
+using commands like the following:
+
+    npm install -g browserify
+    browserify -r ajv -o ajv.bundle.js
```
amb26: Amazing - does this actually work?
Together with gpii-json-schema and infusion too?

src/js/server/schemaHandler.js
```diff
@@ -0,0 +1,46 @@
...
+// 1. `options.schemaKey`: A short key for the schema that will represent it in the `Content-Type` header.
+// 2. `options.schemaUrl`:  A URL for the schema, which will be included in both the `Link` and `Content-Type` headers.
+//
+// Note that this will not actually implement the required `handleRequest` invoker.  You are expected to do that in your
+// own component, but you can use `that.sendResponse` as you would normally with any `gpii.express.handler` grade.
+//
+// See the tests in this package for working examples.
+//
+"use strict";
+var fluid = fluid || require("infusion");
```
amb26: client include

tests/js/handler-caseholder.js
```diff
@@ -0,0 +1,61 @@
+// Test `gpii.schema.middleware` and its subcomponents.
+//
+"use strict";
+var fluid        = fluid || require("infusion");
+var gpii         = fluid.registerNamespace("gpii");
+var jqUnit       = require("jqUnit");
+
+require("../../node_modules/gpii-express/tests/js/lib/test-helpers");
```
amb26: These will all break in npm 3. Use module-relative includes instead. It looks like no special trickery is required, and this can simply be 
require("gpii-express/tests/js/lib/test-helpers") etc.

tests/js/parser-server-tests.js
```diff
@@ -0,0 +1,155 @@
...
+var jqUnit = require("jqUnit");
+var path   = require("path");
+var schemaDir = path.resolve(__dirname, "../schemas");
+
+require("../../src/js/common/parser");
+require("../../src/js/common/validate");
+require("../../src/js/server/validate");
+
+// Utility function to wire up and initial wait into each test case.
+require("../../node_modules/gpii-express/tests/js/lib/test-helpers");
```
amb26: See above

tests/js/parser-server-tests.js
```diff
@@ -0,0 +1,155 @@
...
+
+  Test the parser by itself with as little interaction with the validator as possible (only static validator functions used).
+
+  Client-side tests and integration are covered by the validator tests, which make use of the parser internally.
+
+ */
+"use strict";
+var fluid  = fluid || require("infusion");
+var gpii   = fluid.registerNamespace("gpii");
+var jqUnit = require("jqUnit");
```
amb26: Needs to be updated to "node-jqunit" for npm dependency

tests/static/validate-client-tests.html
```diff
@@ -0,0 +1,170 @@
...
+                            "type":    "object",
+                            "definitions": {
+                                "required": {
+                                    "type": "boolean"
+                                },
+                                "boolean": {
+                                    "type": "boolean"
+                                },
+                                "date": {
+                                    "type":   "string",
```
amb26: Spacing has become rather peculiar in these multi-line JSON values. Also, can we break out this giant script into a js include? That way it can be linted properly

README.md
```diff
@@ -1 +1,133 @@
...
+
+  require("gpii-json-schema");
+  var validator = gpii.schema.validator({
+    schemaContents: {
+      sample: { "type": "object", "properties": { "required": { "type": "boolean" } }, "required": ["required"]}
+    }
+  });
+
+  var errors = validator.validate("sample", { foo: "bar" });
+  if (errors) {
```
amb26: You need to explain what the layout of the errors block is here in detail.

README.md
```diff
@@ -1 +1,133 @@
...
+
+# Adding headers to REST responses
+
+When writing REST interfaces, we also commonly return JSON data in response to a request.  JSON Schemas can also be
+used to provide hints about the output format we are using.
+
+The working group that writes the JSON Schema standard [has outlined two approaches for labeling outgoing responses](http://json-schema.org/latest/json-schema-core.html#anchor33).
+Both of these involve setting HTTP headers in the outgoing response, as in:
+
+    Content-Type: application/my-media-type+json; profile="http://example.com/my-hyper-schema#"
```
amb26: Any ideas why they call this a "hyper-schema"? It looks like it ought to be just a regular fucking schema to me.

package.json
```diff
@@ -0,0 +1,34 @@
...
+    "grunt": "~0.4.4",
+    "grunt-contrib-jshint": "~0.9.0",
+    "grunt-gpii": "git://github.com/GPII/grunt-gpii.git#58ae055833f437a9e91baedb9f39a22c5fdd727f",
+    "grunt-jsonlint": "1.0.4",
+    "grunt-shell": "0.6.4",
+    "infusion": "git://github.com/fluid-project/infusion.git#282f1a318718eed0b0ec060fb8b4ad254417fd7e",
+    "json-schema-ref-parser": "1.4.0"
+  },
+  "devDependencies": {
+    "gpii-express": "git://github.com/the-t-in-rtf/gpii-express.git#8730441601db239870046c796b85c0f020a302a1",
```
amb26: It looks like this is not actually a devDependency, but actually a core-dependency, because of schemaHandler and schemaMiddleware, right?

src/js/common/parser.js
```diff
@@ -0,0 +1,223 @@
+/*
```
amb26: This large comment needs to go into some proper docs. I am thinking that we need a more structured summary of the errors rather than simply reporting the human-readable message. The user will need some means of determining, somehow, WHICH validation rule, in terms of the schema, was responsible for the message.

src/js/common/parser.js
```diff
@@ -0,0 +1,223 @@
...
+  rules.  Even a simple validation error for a field that uses `allOf` and `anyOf` would return at least two warnings
+  from the validator, one of which simply says how many rules failed, or that one of them must match.
+
+  If the desired schema property lacks a description, the raw message from the validator will be used as a fallback.
+
+  To use this component, you will need to instantiate it and make it aware of your schemas.  These will be dereferenced
+  and cached as they are added.  You can then use the `lookupField` method to retrieve any property of the field from
+  the schema definition, as in:
+
+    parser.lookupField("user", ".password.description")
```
amb26: What will it return, in response to this call?

A good option would be to produce MD docs in the style of the new core framework docs, e.g.

https://github.com/amb26/infusion-docs/blob/FLUID-5791/src/documents/CoreAPI.md#fluidiotan-first

src/js/common/parser.js
```diff
@@ -0,0 +1,223 @@
...
+  Note that this library expects to be able to resolve external references relative to the first argument passed to its
+  `dereference` function, and it expects to load and cache the references itself as needed.  Since we are receiving keys
+  from the `validator`, we need to some how convert these to URLs.  We do this using `fluid.stringTemplate` to combine
+  `options.schemaDir` with the individual key, which means:
+
+  1.  If `options.schemaDir` is a directory location, a full path will be produced.
+  2.  If `options.schemaDir` is a URL, a full URL will be produced.
+  3.  If `options.schemaDir` is empty, paths will be resolved relative to the working directory.
+
+  If you are using this with the `validator`, it takes care of that bit of wiring for you.
```
amb26: What does the configuration look like, if I am using this together with the validator?

src/js/common/parser.js
```diff
@@ -0,0 +1,223 @@
...
+
+  To use this component, you will need to instantiate it and make it aware of your schemas.  These will be dereferenced
+  and cached as they are added.  You can then use the `lookupField` method to retrieve any property of the field from
+  the schema definition, as in:
+
+    parser.lookupField("user", ".password.description")
+
+  This component uses [json-schema-ref-parser](https://github.com/BigstickCarpet/json-schema-ref-parser).  We configure
+  it using `options.parserOptions` (empty by default).  Read the library's documentation for details of what options
+  are available.
```
amb26: Please link directly to their docs page describing these

README.md
colinbdclark: Hyper-pastry! :cake:

src/js/common/parser.js
```diff
@@ -0,0 +1,223 @@
...
+            }
+            else {
+                return false;
+            }
+        }
+
+        return currentLevel;
+    }
+
+    // If we can't evolve the output, return `false` so that we can use the existing error.
```
amb26: It's not clear what the "existing error" is, given the definition of this function

src/js/common/parser.js
```diff
@@ -0,0 +1,223 @@
...
+gpii.schema.parser.updateSchemas = function (that) {
+    var promises = [];
+    fluid.each(that.model.schemas, function (schemaContent, schemaKey) {
+        if (!that.dereferencedSchemas[schemaKey]) {
+            promises.push(gpii.schema.parser.dereference(that, schemaKey));
+        }
+    });
+
+    if (promises.length > 0) {
+        fluid.promise.sequence(promises).then(function () {
```
amb26: You have no rejection handler here! Extremely shady!

src/js/common/validate.js
```diff
@@ -0,0 +1,343 @@
...
+        // Errors are associated with the right field based on the `dataPath` received from AJV.
+        var path = gpii.schema.validator.extractPathSegmentsFromError(error);
+
+        var errorMessage      = error.message;
+        var overwriteExisting = false;
+
+        // If we have a parser, we will evolve the output if possible, and use that to replace every raw message for
+        // the same field.
+        //
+        if (that.parser) {
```
amb26: Why might we not have a parser?

tests/js/validate-function-tests.js
```diff
@@ -0,0 +1,133 @@
...
+
+jqUnit.test("Test sanitizing path segment with square brackets...", function () {
+    var output = gpii.schema.validator.sanitizePathSegment(".['this.works']");
+    jqUnit.assertEquals("The output should be as expected...", "this.works", output);
+});
+
+
+jqUnit.test("Extract the path for a missing required top-level field....", function () {
+    var error    = { keyword: "required", dataPath: ".required", message: "message" };
+    var path     = gpii.schema.validator.extractPathSegmentsFromError(error);
```
amb26: These 9 tests are just copypasta and should be consolidated into a declarative form

src/js/common/validate.js
```diff
@@ -0,0 +1,343 @@
...
+
+ */
+
+"use strict";
+var fluid = fluid || require("infusion");
+var gpii  = fluid.registerNamespace("gpii");
+
+var Ajv = require("ajv");
+
+fluid.registerNamespace("gpii.schema.validator");
```
amb26: I think that all of this file would be profitably namespaced into something reflecting the particular technology we are using, where we are doing something which depends on it (e.g. parsing its strange format for referring to data paths etc.) - perhaps gpii.schema.validator.ajv

Anything which simply depends on the ref parser (if there is any such stuff) can remain in top-level gpii.schema.validator or gpii.schema

src/js/common/validate.js
```diff
@@ -0,0 +1,343 @@
...
+//
+// If you try to resolve a path that does not exist and `createMissingSegments` is not set, `undefined` will be returned.
+//
+gpii.schema.validator.resolveOrCreateTargetFromPath = function (target, path, createMissingSegments) {
+    var resolvedTarget = target;
+    for (var a = 0; a < path.length; a++) {
+        var segment = path[a];
+        if (!resolvedTarget[segment]) {
+            if (createMissingSegments) {
+                resolvedTarget[segment] = a === path.length - 1 ? [] : {};
```
amb26: If it weren't for this strange rule about how to fill in missing elements, this could just be replaced by fluid.get or fluid.getSimple - why do we have this rule?

src/js/common/validate.js
```diff
@@ -0,0 +1,343 @@
...
+gpii.schema.validator.sanitizePathSegment = function (segment) {
+    // Discard any leading dot
+    var segmentMinusLeadingDot = segment.replace(/^\./, "");
+
+    // If we are surrounded by `['']`, extract the inner content and then unescape it.
+    var hasSpecialRegexp = /\['(.+)'\]/;
+    var specialMatches = segmentMinusLeadingDot.match(hasSpecialRegexp);
+    if (specialMatches) {
+        // unescape the remaining variables by using them to create a new `String`.
+        return String(specialMatches[1]);
```
amb26: What type might it be otherwise?

src/js/server/schemaMiddleware.js
```diff
@@ -0,0 +1,138 @@
...
+            funcName: "gpii.schema.middleware.rejectOrForward",
+            args:     ["{that}", "{arguments}.0", "{arguments}.1", "{arguments}.2"]
+        }
+    },
+    // Approach adapted from `gpii.express.requestAware.router`.
+    events: {
+        "onInvalidRequest": null
+    },
+    dynamicComponents: {
+        requestHandler: {
```
amb26: Dubious - why do we duplicate this material from the router? It shouldn't be necessary for something which acts a simple middleware to construct request components

tests/js/parser-server-tests.js
```diff
@@ -0,0 +1,155 @@
...
+
+gpii.schema.parser.tests.server.validateAndTest = function (validator, schemaKey, content, expected) {
+    var output = validator.validate(schemaKey, content);
+    jqUnit.assertDeepEq("The output should be as expected...", expected, output);
+};
+
+
+fluid.defaults("gpii.schema.parser.tests.server.caseHolder", {
+    gradeNames: ["gpii.express.tests.caseHolder"],
+    sequenceStart: [
```
amb26: We should be able to economise on this sequence start material by using a standard testCaseHolder from gpii-express - subject to relying on particularly named events in our own holder

tests/js/parser-server-tests.js
```diff
@@ -0,0 +1,155 @@
...
+                    validator: {
+                        type: "gpii.schema.validator.server.hasParser",
+                        options: {
+                            schemaDir: schemaDir,
+                            components: {
+                                parser: {
+                                    options: {
+                                        listeners: {
+                                            "onSchemasUpdated.notifyEnvironment": {
+                                                func: "{testEnvironment}.events.onSchemasUpdated.fire"
```
amb26: This vast tree can just become one distributeOptions

tests/schemas/base.json
```diff
@@ -0,0 +1,60 @@
...
+            "type":     "number",
+            "minValue": 1,
+            "maxValue": 10
+        },
+        "array": {
+            "type": "array",
+            "items": {"type": "boolean" }
+        },
+        "regex": {
+            "description": "The string should be five characters long, begin with 'v' and end with 'd'.",
```
amb26: Doesn't it also ensure that the middle character is "l"?

README.md
the-t-in-rtf: Too much coffee?  The hyper schema working group must have worked on the docs?  Anyway, we don't require any of the hyper-ness yet, I can update the example to avoid even an indirect suggestion that it is preferred.

.jshintrc
the-t-in-rtf: Done.

README.md
the-t-in-rtf: Done.

README.md
the-t-in-rtf: Yes, the client-side tests use a browserified version of ajv prepared in this manner.

README.md
the-t-in-rtf: Done.

README.md
the-t-in-rtf: Done.

src/js/server/schemaHandler.js
the-t-in-rtf: Done, throughout the server-side component code and tests.

tests/js/handler-caseholder.js
the-t-in-rtf: I followed your advice from a previous pull request and replaced this with a call to `require("gpii-express");` followed by `gpii.express.loadTestingSupport()`.

tests/js/parser-server-tests.js
the-t-in-rtf: See above... ;)

tests/js/parser-server-tests.js
the-t-in-rtf: Yes, I had to do this throughout to get everything working in the post-dedupe infusion world.

README.md
the-t-in-rtf: There is alternate wording here now.

tests/static/validate-client-tests.html
the-t-in-rtf: Moved to a separate include and reindented per the standard format.

package.json
the-t-in-rtf: Good catch, I was only using it for testing before adding the gatekeeper modules.

the-t-in-rtf: Done.

tests/schemas/base.json
the-t-in-rtf: Updated the regexp to match the description.

src/js/common/parser.js
the-t-in-rtf: What specific user would need to be able to look at a rule in our JSON Schemas?  My assumption was the opposite, that end users shouldn't even be aware that we're using JSON Schemas at all.

src/js/common/parser.js
the-t-in-rtf: Done.

src/js/common/parser.js
amb26: Eventually we will have USER TOOLS. That is, in response to some diagnostic from the system, they should be guided, visually, towards an explanation of what caused it and what they can do to make it go away. The end users may or may not be aware "that we're using JSON Schemas at all" but hopefully they will be aware that we have built a system that does more than just push diagnostics at them in the form of textual messages. Whilst this day is some way off, we should do our best to consciously build scaffolding for it.

src/js/common/validate.js
the-t-in-rtf: The parser was added later in the process, it was convenient to have the ability to work without it while I was migrating.  I have cleaned this up and collapsed the `hasParser` grades into their respective defaults.

src/js/common/parser.js
the-t-in-rtf: We should talk offline and record some notes here.  I am already using this with front end components, and am already using the scaffolding I consciously built to provide meaningful feedback (though textual) in the right context.

src/js/common/parser.js
the-t-in-rtf: I have added a function to pass any errors to `fluid.fail`.  I could use your help in getting the tests working for this.

the-t-in-rtf: https://github.com/the-t-in-rtf/gpii-json-schema/commit/d9e49e2c23f72c8c461dee9e064ab26a1d0ce84a

src/js/common/validate.js
the-t-in-rtf: We are creating deep structures that mimic the structure of the original JSON.  Let's say our JSON looks like:

```
{ foo: { bar: { baz: "bogus" } }
```

For each error, we are given a path by the validator, something like `.foo.bar.baz`.  We want to end up with something that closely matches the structure of the original JSON, more like:

```
{
  foo: {
    bar: {
      baz: ["that's not a valid option"]
    }
  }
}
```

To produce this, we use the path information returned by the validator, which may refer to a quite deep error.  This code creates the intermediate structures.  If there is a better way to do this, I would love to hear it.

amb26: It seems to me that simply 

var value = fluid.get(target, path);
if (!value && createMissingSegments) {
    value = [];
    fluid.set(target, path, value);
}
return value;

is equivalent and rather clearer

src/js/common/parser.js
the-t-in-rtf: We chatted briefly, really we are just talking about exposing the path to the error in the schema in addition to the description, so that it is possible to wire in an alternate feedback system.  That system would be expected to make sense of the schema and the path provided on its own.

I also have been thinking about adding a grade that helps associate the feedback with onscreen elements using a system of bindings and/or rules as done in previous work.

src/js/common/validate.js
the-t-in-rtf: Agreed, and done.

src/js/server/schemaMiddleware.js
the-t-in-rtf: I gave this some thought, I agree that we should have a way to call the standard "send" functions without requiring all of the overhead that full `gpii.express.handler` grades have.  I had to begin by adding a simpler "base" handler grade that could be used in this package.  [That PR](https://github.com/GPII/gpii-express/pull/18) is ready for your review.

the-t-in-rtf: I have successfully updated this module to handle rejected requests using the standard handler function without dynamically creating a request handler each time.  We will have to review other uses of the "handler" infrastructure in future pull requests, it may make sense to convert others to use the simpler "base" implementation.

tests/js/parser-server-tests.js
the-t-in-rtf: Done.

tests/js/parser-server-tests.js
the-t-in-rtf: I collapsed the use of the gatekeeper by one level and used `distributeOptions` for the deep options setting.  I left the `schemaDir` option alone, as the syntax to use `distributeOptions` would have been longer in that case.

src/js/common/validate.js
```diff
@@ -0,0 +1,326 @@
...
+    // 1. `simple.simon`
+    // 2. `['kinda.complex'].simple`
+    // 3. `simple['kinda.complex']`
+    // 4. `['really.complex']['don\'t I know it']`
+    //
+    // See https://xkcd.com/1171/
+    //
+    // The matching within single quotes must be non-greedy to ensure that things like `['a.b']['c.d']` are treated as
+    // separate segments.
+    var slicingRegexp = /^\.?(\['.+?']|[^\[\.]+)([\[\.].+)$/;
```
amb26: Doesn't look like it handles double quotes, as advertised in the comments

src/js/common/validate.js
```diff
@@ -0,0 +1,326 @@
...
+  This function takes care of reversing this process including all combinations, to ensure that we end up with the
+  original unescaped literal key.
+
+ */
+gpii.schema.validator.sanitizePathSegment = function (segment) {
+    // Discard any leading dot
+    var segmentMinusLeadingDot = segment.replace(/^\./, "");
+
+    // If we are surrounded by `['']`, extract the inner content and then unescape it.
+    var hasSpecialRegexp = /\['(.+)'\]/;
```
amb26: See above -- shouldn't there be a relationship between this regexp and the one above?

tests/js/middleware-tests.js
```diff
@@ -0,0 +1,75 @@
+/* Tests for the "express" and "router" module */
+"use strict";
+var fluid =  require("infusion");
+var gpii  = fluid.registerNamespace("gpii");
+
+var path = require("path");
+var schemaDir = path.resolve(__dirname, "../schemas");
```
amb26: Update this to the automatic scheme that we use in express and elsewhere

tests/js/parser-update-failure-tests.js
```diff
@@ -0,0 +1,39 @@
...
+        "onCreate.misbehave": {
+            funcName: "fluid.fail",
+            args: ["DIE!"]
+        }
+    }
+});
+
+jqUnit.test("Testing parser resolutions of failed promise...", function () {
+    jqUnit.expectFrameworkDiagnostic("The component should thrown an error on startup...", gpii.schema.tests.validator.bornToDie, ["ENOENT"]);
+    //jqUnit.expectFrameworkDiagnostic("The component should thrown an error on startup...", gpii.fuck.off, ["DIE!"]);
```
amb26: Why is this one commented out - was this just a diagnostic attempt?

amb26:
Re the 5000 lines of tests/static/js/ajv.bundle.js - do we need to have this committed, or couldn't we generate it using a postinstall build task?

tests/js/validate-function-tests.js
the-t-in-rtf: I consolidated all of the tests to use a single runner function and a declarative form.

src/js/common/parser.js
the-t-in-rtf: I am thinking this through.  To avoid reproducing a lot of boilerplate (especially for involved concepts), I was thinking of simply linking to the documentation in a comment above each function.

src/js/common/parser.js
the-t-in-rtf: I update the comment to clarify that `false` is returned if the field cannot be looked up, and that it is the responsibility of code that uses this function to respond to take action when a field cannot be looked up.

src/js/common/validate.js
the-t-in-rtf: It is all specific to AJV, and it's fair to namespace it that way.  If we have to change again we can talk about abstracting out some kind of general contract at a higher namespace.

src/js/common/validate.js
the-t-in-rtf: This may have been a bit of voodoo left over from a previous implementation, all tests pass if we simply return the inner matched value, so I have made that change.

tests/js/parser-update-failure-tests.js
the-t-in-rtf: Apparently I didn't finish getting it working, I meant to add a TODO there to talk with you about it.  Mission inadvertently accomplished?  Jeez, I have a potty mouth.

the-t-in-rtf: Added the test file to `all-tests.js` so that my shame will at least be appropriately public until we fix it.

src/js/common/parser.js
the-t-in-rtf: @amb26, I can confirm manually that the promise rejection is functioning here, but cannot get the tests in `parser-update-failure-tests.js` to work.  I could use some advice.

src/js/common/parser.js
the-t-in-rtf: I am having difficulty cleanly linking between the Javascript and the associated docs, but you can at least see the work in progress here:  https://github.com/the-t-in-rtf/gpii-json-schema/blob/GPII-1336/docs/parser.md

src/js/common/parser.js
the-t-in-rtf: This is no longer relevant as we only have components that work with a parser.

the-t-in-rtf:
As this is getting a bit long, I wanted to summarize what's left:

1. Reviewing the new documentation format.
2. Updating the error map format to include more information about the failing rule.

I also still want to create a client-side component that uses the `gpii-handlebars` binder and renderer to display summary and inline feedback, but I am debating whether to do that now or evolve a proof-of-concept in working on another package.

the-t-in-rtf:
@amb26 left a top-level comment above about removing the static "browserified" version of AJV from our codebase.

I experimented briefly with `grunt-browserify`, which would have avoided having to install `browserify` globally.  However, the "alias" syntax required to tell `browserify` to expose the correct top-level `Ajv` variable did not seem to work, and `grunt-browserify` in general seems somewhat abandoned.

I discovered that that AJV team have already provided a bundling task as part of their `package.json`, which presumably is part of their browser testing. I used the `grunt-exec` plugin to launch their bundling task . This requires `browserify` to be installed globally, but is otherwise reasonable.

I have updated the README to reflect the new process and the requirement to have `browserify` installed globally.

The new process works on both OS X and Windows and the legacy checked-in copy of AJV has been removed.

docs/parser.md
```diff
@@ -0,0 +1,247 @@
...
+ using the parser component's `dereferenceSchema` invoker and the `schemaKey` attribute.
+
+## gpii.schema.parser.evolveError(that, schemaKey, error)
+
+* `that`: The parser component itself.
+* `schemaKey {String}`: The `key` of the schema (generally the filename, i.e. `base.json`).
+* `error {object}`: The original error returned by AJV.
+* Returns: `Object` The "evolved" error with an updated message based on the `errors` definitions (if there are any).
+
+This function used by the validator to transform the full set of raw validation results using `fluid.transform`.  It
```
amb26: There should be a clearer definition here of exactly what fields are added to the supplied error, their types and meaning. Also, this function should be clear on whether it copies the supplied "error" or modifies it in place.

docs/parser.md
```diff
@@ -0,0 +1,247 @@
...
+# Functions
+
+## gpii.schema.parser.dereferenceSchema(that, schemaPath, schemaKey)
+
+ * `that`: The parser component itself.
+ * `schemaPath {String}`:  A path or URI to the directory or base URL where the schema can be found.
+ * `schemaKey {String}`: The filename of the schema relative to `schemaPath`.
+ * Returns: A `promise` that will be satisfied when the parser finishes its work.
+
+ Dereference a single schema.  As the process is asynchronous, this function returns a promise.  Typically accessed
```
amb26: Should be clear on whether the promise resolves to a value or not (it appears that it doesn't)

docs/parser.md
```diff
@@ -0,0 +1,247 @@
...
+## `gpii.schema.parser.server`
+
+A server-side version of the component which can handle package-relative `schemaPath` entries like `%gpii-package-name/path/within/package`.
+
+# Functions
+
+## gpii.schema.parser.dereferenceSchema(that, schemaPath, schemaKey)
+
+ * `that`: The parser component itself.
+ * `schemaPath {String}`:  A path or URI to the directory or base URL where the schema can be found.
```
amb26: This use of "URI" is unclear - if this URI does not designate a means of access to the schema (a URL), how is it to be used?

docs/parser.md
```diff
@@ -0,0 +1,247 @@
...
+using `errors` at the document level, taken from the v5 draft proposal:
+
+```
+{
+  "properties": {
+    "age": { "minimum": 13 },
+    "gender": { "enum": ["male", "female"] }
+  },
+  "errors": {
+    "#/properties/age/minimum": "Should be at least ${schema} years, ${data} years is too young.",
```
amb26: Is there a description anywhere of what escaping is intended to be used in these reference?

docs/middleware.md
```diff
@@ -0,0 +1,106 @@
...
+
+The [rules to use in transforming](https://github.com/fluid-project/infusion-docs/blob/0e3862aaab38742c71f8f6e3e155a3b6d5199ad4/src/documents/ModelTransformationAPI.md#fluidmodeltransformwithrulessource-rules-options)
+validation errors before they are sent to the user (see above).
+
+# Functions
+
+## `gpii.schema.middleware.rejectOrForward(that, req, res, next)`
+
+* `that`: The component itself.
+* `req`: The [request object](http://expressjs.com/en/api.html#req) provided by Express.
```
amb26: It's probably useful here to punt even further up the docs chain, and in addition to the link you provide here, to also be clear that what this object actually *IS* is the "http.IncomingMessage" object supplied by the raw node.js http server API: https://nodejs.org/api/http.html#http_class_http_incomingmessage

similarly for "res"

docs/parser.md
```diff
@@ -0,0 +1,247 @@
...
+
+The base component is designed to work both on the client and server side.
+
+## `gpii.schema.parser.server`
+
+A server-side version of the component which can handle package-relative `schemaPath` entries like `%gpii-package-name/path/within/package`.
+
+# Functions
+
+## gpii.schema.parser.dereferenceSchema(that, schemaPath, schemaKey)
```
amb26: These are better advertised as component-bound facilities (via an invoker) rather than concrete public functions, since they accept a component as the first argument. Advertising them as component-bound gives the possibility for variant implementations, conforming to an abstract contract.

However, you do need to document the full options structure accepted by the parser's grade.

docs/parser.md
```diff
@@ -0,0 +1,247 @@
...
+}
+```
+
+# Requirements
+
+To use this component, you will need to instantiate it and make it aware of your schemas.  These will be dereferenced
+and cached as they are added.  You can then use the `evolveError` method to resolve the custom error definition for a
+field using the schema key and the field's JSON pointer.  See the validator implementation for an example of its use.
+
+This component uses [json-schema-ref-parser](https://github.com/BigstickCarpet/json-schema-ref-parser).  We configure
```
amb26: This documentation should be organised as a full table of all options accepted by the component - that is, with columns, "member, type, description" as, for example, at

http://docs.fluidproject.org/infusion/development/ContextAwareness.html#-checkrecord-members-in-an-adaptationrecord-

docs/middleware.md
```diff
@@ -0,0 +1,106 @@
...
+the incoming data before validation (see above).
+
+## `options.rules.validationErrorsToResponse`
+
+The [rules to use in transforming](https://github.com/fluid-project/infusion-docs/blob/0e3862aaab38742c71f8f6e3e155a3b6d5199ad4/src/documents/ModelTransformationAPI.md#fluidmodeltransformwithrulessource-rules-options)
+validation errors before they are sent to the user (see above).
+
+# Functions
+
+## `gpii.schema.middleware.rejectOrForward(that, req, res, next)`
```
amb26: This should be packaged as a piece of middleware in itself - so that the user does not need to go to an extra step of defining the actual middleware function themselves (by dispatching on the "that" argument) in order to actually use it as middleware. I suggest a member named "middleware" produced on the component, so that "that.middleware" can simply be contributed into a standard express pipeline

docs/middleware.md
```diff
@@ -0,0 +1,106 @@
...
+The path (URI or filesystem) to a directory that contains a file matching `options.schemaKey`.
+
+## `options.responseSchemaKey`
+The schema key that [our handler](./handler.md) will use in constructing response headers.
+
+## `options.responseSchemaUrl`
+A base URL where `options.responseSchemaKey` can be found.
+
+## `options.rules.requestContentToValidate`
+The [rules to use in transforming](https://github.com/fluid-project/infusion-docs/blob/0e3862aaab38742c71f8f6e3e155a3b6d5199ad4/src/documents/ModelTransformationAPI.md#fluidmodeltransformwithrulessource-rules-options)
```
amb26: Make this link to our published docs rather than the raw markdown (as well as for all other such links). In this case, the corresponding link is http://docs.fluidproject.org/infusion/development/ModelTransformationAPI.html#fluid-model-transformwithrules-source-rules-options-

docs/validator.md
```diff
@@ -0,0 +1,261 @@
...
+value passed to `validate` is expected to match one of the keys in `options.schemaContent`.
+
+You can change the behavior of AJV by updating `options.validatorOptions` (check their documentation for the syntax).
+
+# Reusing Content between Schemas
+
+One of the key strengths of JSON Schema is that it allows you to compose a complex schema out of parts taken from
+other schemas.  The validator in this package supports references between JSON Schemas.
+
+Inheritance is still a sticking point at least in v4 of the draft standard.   You cannot safely expect to overlay
```
amb26: Can you give an example here to exhibit what you mean by "reuse individual definitions between schemas"?

docs/validator.md
```diff
@@ -0,0 +1,261 @@
...
+
+Transform raw validator output into a more human-readable form that corresponds to the structure of the original JSON
+input.  Typically accessed using the validator component's `validate` invoker and the last two arguments.
+
+
+## `gpii.schema.validator.ajv.sanitizeValidationErrors(that, schemaKey, errors)`
+
+* `that`: The component itself.
+* `schemaKey {String}`: The schema key we are validating against.
+* `errors`: A map of existing error results to be sanitized.
```
amb26: A map of what keys to what values?

src/js/client/errorBinder.js
```diff
@@ -0,0 +1,145 @@
...
+
+/* global fluid, $ */
+(function () {
+    "use strict";
+    var gpii = fluid.registerNamespace("gpii");
+    fluid.registerNamespace("gpii.schemas.client.errorBinder");
+
+    gpii.schemas.client.errorBinder.displayErrors = function (that) {
+        // Get rid of any previous validation errors.
+        fluid.each(that.locate("validationError"), function (element) {
```
amb26: I believe that, by jQuery standard semantics, that.locate("validationError").remove() is sufficient here

src/js/common/parser.js
```diff
@@ -0,0 +1,188 @@
...
+//
+// https://github.com/the-t-in-rtf/gpii-json-schema/blob/GPII-1336/docs/parser.md#gpiischemaparserresolvejsonpointerthat-schemakey-schemafieldpath
+//
+gpii.schema.parser.resolveJsonPointer = function (that, schemaKey, jsonPointer) {
+    var dereferencedSchema = that.dereferencedSchemas[schemaKey];
+    return jsonpointer.get(dereferencedSchema, jsonPointer);
+};
+
+
+// Look up the error metadata for a single field. See the documentation for details:
```
amb26: Sadly I fear that until we have a docs site set up, the docs comments here will need to duplicate the material which is in the .md file. This is what we do in core Infusion as well - readers of the code need to be able to read it in a self-contained way. Also, you already have a cut and paste burden since all of these links will change once the code is committed.

src/js/server/parser.js
```diff
@@ -0,0 +1,27 @@
+/*
+
+    The server-side version of the parser.  The only difference is that on the server side, we can resolve relative
+    paths using `fluid.module.resolvePath`.
+
+  */
+
+"use strict";
+var fluid = fluid || require("infusion");
```
amb26: Server-style include here

src/js/server/schemaHandler.js
```diff
@@ -0,0 +1,53 @@
...
+var fluid = require("infusion");
+var gpii  = fluid.registerNamespace("gpii");
+
+require("gpii-express");
+
+fluid.registerNamespace("gpii.schema.handler");
+
+// Send the appropriate headers and then let the underlying grade's `sendResponse` function take over.
+gpii.schema.handler.sendResponse = function (that, response, statusCode, body) {
+    if (!that.options.schemaKey || !that.options.schemaUrl) {
```
amb26: Use some kind of "expect" utility to specifically report which field is missing

src/js/server/schemaHandler.js
```diff
@@ -0,0 +1,53 @@
...
+        fluid.log("Your gpii.schema.handler is not configured correctly and cannot set the appropriate headers.");
+    }
+    else if (response.headersSent) {
+        fluid.log("Can't set headers, they have already been sent.");
+    }
+    else {
+        response.type("application/" + that.options.schemaKey + "+json; profile=\"" + that.options.schemaUrl + "\"");
```
amb26: No escaping rules here? What characters are forbidden from appearing in a "schemaKey" by virtue of its use in a literal header here?

src/js/server/schemaMiddleware.js
```diff
@@ -0,0 +1,186 @@
...
+        handleRequest: {
+            func: "{that}.sendResponse",
+            args: ["{that}.options.statusCode", "{that}.options.message"]
+        }
+    }
+});
+
+fluid.registerNamespace("gpii.schema.middleware");
+
+gpii.schema.middleware.rejectOrForward  = function (that, req, res, next) {
```
amb26: See above comment - to be helpful to users, this should actually conform to the middleware contract

docs/middleware.md
amb26: OK - this is a clear case where our decision in favour of documenting component-bound material rather than the raw global functions is clearly correct. I missed when reading this the first time that the component does *indeed* host a member (invoker) "that.middleware" which conforms to the middleware contract - and it is that member which should be documented

src/js/server/parser.js
the-t-in-rtf: Fixed.

docs/parser.md
the-t-in-rtf: There was a little bit on the cloned return value further down, which I moved up and expanded.

docs/parser.md
the-t-in-rtf: Done.

docs/parser.md
the-t-in-rtf: Done for both the `parser` and `validator`.

docs/middleware.md
the-t-in-rtf: Done.

docs/middleware.md
the-t-in-rtf: Done.

docs/parser.md
the-t-in-rtf: Done.

docs/parser.md
the-t-in-rtf: I expanded this a little, take a look.

docs/parser.md
the-t-in-rtf: It's part of the JSON pointer standard, I added a short block explaining the two escaping conventions and linking to the standard.

docs/middleware.md
the-t-in-rtf: Done for both.

docs/parser.md
the-t-in-rtf: Done on both counts.

docs/parser.md
the-t-in-rtf: Done.

src/js/common/parser.js
the-t-in-rtf: This is not merely a cut and paste burden, as we have invoker-focused language in the docs and would need to figure out how to present that next to the full functions.  We also have a large problem in that markdown is not at all rendered in comment either in IDEs or on GitHub.  Even links to external documentation are not supported.  We would need two formats for options blocks, two formats for external links, and much more.

The division between the inline and external documentation is not something docpad will solve, either.  I did experiment briefly with JSDoc, but that also lacks markdown support.

As an interim solution I would propose that you and I agree on a minimum useful inline documentation and retain the links to the external docs for more information.  Said links to be updated after the initial merge in a very small PR.

src/js/server/schemaHandler.js
the-t-in-rtf: I have created a new version of the `hasRequiredOptions` grade from `gpii-handlebars`, which is more at home here.  The previous grade was client-only, the new one is usable for both.  I have updated all of the grades with required options to use the new mechanism.  I also added minimal documentation for this, but would prefer to keep it out of the full documentation for the time being.

docs/validator.md
the-t-in-rtf: I updated this with a simple example.

src/js/server/schemaHandler.js
the-t-in-rtf: Looking at the [relevant sections of RFC 1341](https://www.w3.org/Protocols/rfc1341/4_Content-Type.html), we see

```
Content-Type := type "/" subtype *[";" parameter] 

type :=          "application"     / "audio" 
          / "image"           / "message" 
          / "multipart"  / "text" 
          / "video"           / x-token 

x-token := <The two characters "X-" followed, with no 
           intervening white space, by any token> 

subtype := token 

parameter := attribute "=" value 

attribute := token 

value := token / quoted-string 

token := 1*<any CHAR except SPACE, CTLs, or tspecials> 

tspecials :=  "(" / ")" / "<" / ">" / "@"  ; Must be in 
           /  "," / ";" / ":" / "\" / <">  ; quoted-string, 
           /  "/" / "[" / "]" / "?" / "."  ; to use within 
```

As we commonly use periods in our ids/filenames, we are already breaking the convention.  I am looking at libraries for [quote encoding](http://www.ietf.org/rfc/rfc2616.txt) the string now.

the-t-in-rtf: I used [quoted-printable](https://github.com/mathiasbynens/quoted-printable) to wrap the schemaKey content for use in the header.

src/js/client/errorBinder.js
the-t-in-rtf: Done and tested manually.  Updating the validator tests at the moment.

the-t-in-rtf:
As far as I can see, the last remaining work from my end is three-fold:

1. Finish the client-side validator and tests.
2. Move newer blocks of comments to the docs and review those following the recent cleanup.
3. Add summaries with links plus JSDocs hints to the code itself.

src/js/server/schemaHandler.js
```diff
@@ -0,0 +1,70 @@
...
+fluid.defaults("gpii.schema.handler.common", {
+    gradeNames: ["gpii.hasRequiredOptions"],
+    requiredFields: {
+        "schemaKey": true,
+        "schemaUrl": true
+    }
+});
+
+// A companion grade designed for use with `gpii.express.base`.  Intended for static rather than dynamic use.
+fluid.defaults("gpii.schema.handler.base", {
```
amb26: There should be just one grade here - the user shouldn't have to choose between these. There isn't a "static use" in any case since the use cases are the same - it's just a matter of appropriate factoring

docs/middleware.md
```diff
@@ -0,0 +1,102 @@
...
+# Invokers
+
+## `{middleware}.middleware(req, res, next)`
+
+* `req`: The [request object](http://expressjs.com/en/api.html#req) provided by Express, which wraps node's [`http.incomingMessage`](https://nodejs.org/api/http.html#http_class_http_incomingmessage).
+* `res`: The [response object](http://expressjs.com/en/api.html#res) provided by Express, which wraps node's [`http.ServerResponse`](https://nodejs.org/api/http.html#http_class_http_serverresponse).
+* `next`: The next Express middleware or router function in the chain.
+* Returns: Nothing.
+
+This invoker fulfills the standard contract for a `gpii.express.middleware` component.  It examines the `req` content
```
amb26: If it interrupts the conversation, how does it do this? Shouldn't it do this via the 4th "err" argument?

the-t-in-rtf: In my reading, that method signature is for middleware that _responds_ to errors, and not for middleware that _generates_ errors:

http://expressjs.com/en/guide/using-middleware.html#middleware.error-handling

Our middleware interrupts the conversation by neglecting to call the `next` function it is passed, as outlined (in reverse) in [the documentation](http://expressjs.com/en/guide/writing-middleware.html):

> If the current middleware function does not end the request-response cycle, it must call next() to pass control to the next middleware function. Otherwise, the request will be left hanging.

Having our middleware call `next(error)` is also an option here.  It would however mean that we'd have to wire in our own error-handling middleware and ensure that it is always "downstream" of this middleware.  Otherwise we would not have the same level of control over the response sent to the user.

I would suggest keeping the current implementation the way it is unless we have another use case for writing our own error-handling middleware or some more compelling reason why simply choosing to end the conversation and respond directly is not acceptable.

src/js/server/schemaHandler.js
the-t-in-rtf: Can you comment on the [`gpii-express` pull](https://github.com/GPII/gpii-express/pull/18) with details about what you mean about "appropriate refactoring"?

tests/js/parser-update-failure-tests.js
```diff
@@ -0,0 +1,26 @@
...
+fluid.defaults("gpii.schema.tests.validator.bornToDie", {
+    gradeNames: ["gpii.schema.validator.ajv.server"],
+    schemaPath:  "%gpii-json-schema/tests/badSchemas"
+});
+
+// The constructor itself does not call fluid.fail (or at least at not in the way expected here).  The failure is
+// thrown by a promise, which itself is created by a listener that responds to the loading of template content during
+// startup.
+// TODO:  Review with Antranig.
+jqUnit.asyncTest("Testing parser resolutions of failed promise...", function () {
```
amb26: Your best bet here is the "kettle.pushInstrumentedErrors" utility, which pattern you can see demonstrated here:

https://github.com/amb26/kettle/blob/KETTLE-32/tests/ErrorTests.js#L178

src/js/server/schemaMiddleware.js
```diff
@@ -0,0 +1,172 @@
...
+            },
+            "fieldErrors": ""
+        }
+    },
+    events: {
+        "onInvalidRequest": null
+    },
+    handlerGrades: ["gpii.schema.middleware.handler"],
+    dynamicComponents: {
+        requestHandler: {
```
amb26: This is another example of router functionality that belongs upstairs in the router. We shouldn't need to construct an entire handler component because we wish to fire an error.

src/js/server/schemaMiddleware.js
```diff
@@ -0,0 +1,172 @@
...
+    gradeNames: ["gpii.express.router.passthrough"],
+    method:     "post",
+    rules: {
+        requestContentToValidate: {
+            "": "body"
+        }
+    },
+    components: {
+        // required middleware that provides `req.body`
+        json: {
```
amb26: This whole grade seems too heavyweight as well as duplicating definitions which belong upstairs. We should have a call to discuss how it can be better factored and become more usable with less dependence on gpii-express router infrastructure.

src/js/server/schemaHandler.js
```diff
@@ -0,0 +1,54 @@
...
+/**
+ *
+ * Send the appropriate headers and then let the underlying grade's `sendResponse` function take over.
+ *
+ * @param that - The handler component itself.
+ * @param response {Object} - The Express `response` object: http://expressjs.com/en/api.html#res
+ * @param statusCode - The numeric HTTP status code.
+ * @param body {Object} - The response payload to send to the browser.
+ */
+gpii.schema.handler.sendResponse = function (that, response, statusCode, body) {
```
amb26: This is too heavyweight and should instead be packaged as a piece of middleware rather than a modified "sendResponse" holder.

src/js/common/parser.js
```diff
@@ -0,0 +1,281 @@
...
+gpii.schema.parser.updateSchemas = function (that) {
+    var promises = [];
+    fluid.each(
+        that.model.schemas, function (schemaContent, schemaKey) {
+        if (!that.dereferencedSchemas[schemaKey]) {
+            promises.push(that.dereferenceSchema(schemaKey));
+        }
+    });
+
+    if (promises.length > 0) {
```
amb26: This length check is unnecessary

src/js/common/parser.js
```diff
@@ -0,0 +1,281 @@
...
+
+/**
+ *
+ * A model listener to dereference and cache a dereferenced version of schemas as they are added to `model.schemas`.
+ */
+gpii.schema.parser.updateSchemas = function (that) {
+    var promises = [];
+    fluid.each(
+        that.model.schemas, function (schemaContent, schemaKey) {
+        if (!that.dereferencedSchemas[schemaKey]) {
```
amb26: I think there is a bug here, in that "dereferencedSchemas" is never cleared, which implies that it can only ever be updated once. You should add a fix and test case for the situation where a schema is fetched once and then updated.

src/js/common/parser.js
```diff
@@ -0,0 +1,281 @@
...
+            funcName: "gpii.schema.parser.resolveJsonPointer",
+            args:     ["{that}", "{arguments}.0", "{arguments}.1"] // schemaKey, jsonPointer
+        },
+        evolveError: {
+            funcName: "gpii.schema.parser.evolveError",
+            args:     ["{that}", "{arguments}.0", "{arguments}.1"] // schemaKey, jsonPointer
+        }
+    },
+    listeners: {
+        "onSchemasUpdated.indicateReadiness": {
```
amb26: I think that this event's semantics are a bit buggy. It can't necessarily be advising the user that the component is *indeed* ready to use since several requests for schema updates may have come in since the last time the flag was set. I think the confusion arises from the use of model-like semantics for the "schemas" which imply that they could be updated at any time after the component is constructed. You could either back this out and make a simpler component that accepts all its schemas at construction time, or you else you will have to go forwards and go the whole hog with a fully dynamic component resolving all the races commented above. If you want to do this, the component will have to maintain a "queue" of currently outstanding requests for schemas which is updated by the modelListener below, "isReady" will have to become a model flag that can become false if the queue is nonempty, etc. and then only become set again once the queue drains successfully.

docs/middleware.md
amb26: OK, I agree with this

src/js/server/schemaHandler.js
the-t-in-rtf: This should just be middleware that adds the two links and lets whomever else finish the conversation.

the-t-in-rtf: I was just thinking about this, the problem comes when we use something like `schemaMiddleware`, which needs to send a response that has the headers set for an error message, and not the headers that would be associated with a successful REST request.  It would either need to call the functions provided here on its own (and duplicate some of the invoker wiring), or would need to pass the request down to its own router that uses the relevant middleware.

src/js/common/parser.js
the-t-in-rtf: Much of the complexity is added by the "dereferencer", which expects to be able to retrieve remote schemas.  I plan to resolve this by moving the dereferencer to the "inline schema" router, which will deliver schema content that has already been dereferenced.

With that moved, the validator will be ready to respond once it has retrieved its schema content, and firing a single `onSchemasLoaded` event will suffice.

the-t-in-rtf:
I finished overhauling the validator and parser.  The client-side validator now works with the schema router, which delivers dereferenced schemas to it.  Thus it no longer uses the client-side dereferencer at all.  The server-side validator and router both use a common server-side component for the dereferencing.  As suggested, schema content is dereferenced once on startup, and thus listening to `onSchemasDereferenced` event is a safe way to tell that we are ready to begin resolving errors.

We should review the open concerns here in our next pairing session.

the-t-in-rtf:
@amb26 This is the next one in line after `gpii-express`.

There are a lot of commits related to upstream updates to `gpii-express`, the biggest recent changes to this package itself are the `validator` and `parser` updates [referenced above](https://github.com/GPII/gpii-json-schema/pull/5#issuecomment-193206032).

the-t-in-rtf:
Apparently I can't leave replies any more.  Anyway, to continue the discussion about [the commented-out tests above](https://github.com/GPII/gpii-json-schema/pull/5#discussion_r54297039):

I followed the suggested approach for the `parser-update-failure-tests.js`.  It still succeeds individually, but fails when run either before or after the rest of the tests.  Please assist.

the-t-in-rtf:
Wrong button.

the-t-in-rtf:
@amb26, I have gone through looking at all of your comments with no reply.  I believe I have addressed all of the open issues.  I uncommented the "lone wolf" test that fails when run in combination with other tests.  You should be able to see the error.  As far as I can see, that's the only outstanding issue here.

tests/js/parser-update-failure-tests.js
the-t-in-rtf: I couldn't get this working even with that method.  I uncommented the broken test so that you can see the problem.

docs/validator.md
the-t-in-rtf: There was an example of the error format immediately below this. I added a "see below" here.

tests/js/middleware-tests.js
the-t-in-rtf: This was fixed a while back.

README.md
the-t-in-rtf: It's explained in the validator docs.

README.md
the-t-in-rtf: This was fixed when I reorganized the docs.

the-t-in-rtf:
The last failing test is now working thanks to recent input from Antranig.  This awaits only the upstream merge of `gpii-express`.

the-t-in-rtf:
The recent changes to the binder required some reworking of the tests.  This is ready to review again.

docs/schemaValidationMiddleware.md
```diff
@@ -0,0 +1,126 @@
...
+          "": "body"
+      }
+
+See `gpii.schema.validationMiddleware.handlesGetMethod` below for an example of working with query data.
+
+### Invokers
+
+#### `{middleware}.middleware(req, res, next)`
+
+* `req`: The [request object](http://expressjs.com/en/api.html#req) provided by Express, which wraps node's [`http.incomingMessage`](https://nodejs.org/api/http.html#http_class_http_incomingmessage).
```
amb26: See comments on other projects - this is not an example of wrapping - the request object *is* the incomingMessage instance

Vagrantfile
```diff
@@ -0,0 +1,131 @@
+# -*- mode: ruby -*-
```
amb26: See comments on other projects re vagrant files - purpose and expected fate of this file should be documented

tests/js/browser/errorBinder-client-tests.js
```diff
@@ -0,0 +1,381 @@
...
+                        func: "{testEnvironment}.browser.exists",
+                        args: [".errorBinder-clientSideValidation-viewport .fieldError"]
+                    },
+                    {
+                        listener: "jqUnit.assertFalse",
+                        event:    "{testEnvironment}.browser.events.onExistsComplete",
+                        args:     ["There should no longer be any field-level errors...", "{arguments}.0"]
+                    }
+                    //,
+                    // TODO: Enable these steps once https://issues.gpii.net/browse/GPII-1580 is resolved.
```
amb26: It seems that GPII-1580 is now resolved - what is the current status of this?

the-t-in-rtf:
As discussed, there is an outstanding TODO: regarding waiting for a client-side fixture to be ready.  We now have a new approach (using executeAsyncScript and the new gpii-webdriver), I will take care of this while pushing through the latest updates this week.

src/js/server/parser.js
```diff
@@ -0,0 +1,111 @@
...
+ * @returns A {Function} that can be passed directly to the parser as a callback.
+ *
+ */
+gpii.schema.parser.getParserCallback = function (that, schemaKey, promise) {
+    return function (error, schema) {
+        if (error) {
+            promise.reject(error);
+        }
+        else {
+            // Our keys have periods in them, which the change applier will interpret as a deeper part of the path
```
amb26: There are a couple of ways round this - either i) supply the path as an array of segments rather than a string, or ii) apply escaping to the path so that . is escaped to \.

the-t-in-rtf: I think GitHub may have eaten your escaping.  Can you describe it perhaps?  SLASH SLASH PERIOD, for example.

the-t-in-rtf: Never mind, I edited your example, you meant escaping as SLASH DOT.

Vagrantfile
the-t-in-rtf: This is now done.

src/js/server/parser.js
the-t-in-rtf: I simplified this to use escaping as discussed.

the-t-in-rtf:
Just to confirm, this is also ready for review, but ideally should be merged after its dependencies, namely:

* gpii-binder
* gpii-express
* gpii-handlebars
* gpii-webdriver

docs/schemaLinkMiddleware.md
```diff
@@ -0,0 +1,77 @@
...
+| ----------- | ---------- | ----------- |
+| `schemaKey` | `{String}` | A short key for the schema that will represent it in the `Content-Type` header. |
+| `schemaUrl` | `{String}` | A URL for the schema, which will be included in both the `Link` and `Content-Type` headers. |
+
+
+## Component Invokers
+
+### `{that}.middleware(request, response, next)`
+
+* `request`: The [request object](http://expressjs.com/en/api.html#req) provided by Express, which wraps node's [`http.incomingMessage`](https://nodejs.org/api/http.html#http_class_http_incomingmessage).
```
amb26: Copypasta remains

docs/schemaLinkMiddleware.md
```diff
@@ -0,0 +1,77 @@
...
+| ------------ | ---------- | ----------- |
+| `statusCode` | `{Number}` |  The [HTTP status code](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) to send via the `response` object. |
+
+
+## Component Invokers
+
+### `{that}.middleware(error, request, response, body)`
+
+* `error`: The error payload returned by upstream middleware.
+* `request`: The [request object](http://expressjs.com/en/api.html#req) provided by Express, which wraps node's [`http.incomingMessage`](https://nodejs.org/api/http.html#http_class_http_incomingmessage).
```
amb26: etc. - perhaps you can just document these in a single place in gpii-express middleware, and just link to it.

docs/schemaValidationMiddleware.md
```diff
@@ -0,0 +1,126 @@
+# The Schema "Gatekeeper" Validation Middleware
+
+Although you usually will build some fault tolerance into your components, on some level they expect to deal with
```
amb26: This apparently value-neutral use of terms like "right", "expected", "invalid", "correct" should be qualified in the first paragraph that you are only prepared to enter into this issue to the extent that the JSON schema standard can process.

docs/validator.md
```diff
@@ -0,0 +1,202 @@
...
+* `errors`: A map of existing error results to be sanitized (see below for examples).
+
+AJV gives us output like:
+
+```
+ [
+     {
+         "keyword": "minLength",
+         "schemaPath": "/password/minLength",
+         "dataPath": ".password",
```
amb26: You should quote or reference whatever the AJV docs say about escaping rules here - or reconstruct it if they have no docs

docs/schemaLinkMiddleware.md
the-t-in-rtf: Done here, and in `gpii-handlebars` and `gpii-pouchdb` as well.

docs/schemaValidationMiddleware.md
the-t-in-rtf: It's a fair point, I have revised the wording to more clearly reflect the limits of this type of validation.

docs/validator.md
the-t-in-rtf: The code is the documentation in this case, so I described the two values that are escaped and linked to the relevant code.

docs/validator.md
```diff
@@ -0,0 +1,205 @@
...
+         "keyword": "required",
+         "schemaPath": "/deep/required/0",
+         "dataPath": ".deep.required",
+         "message": "You must enter a value for the 'deeply required' field."
+     }
+ ]
+```
+See [the documentation for `evolveError`](evolveErrors.md) for more details.
+
+Note that `dataPath` values are ["Javascript property access notation"](https://github.com/epoberezkin/ajv#validation-errors).
```
amb26: Thanks for your researches in discovering this. It appears that this is actually part of a public standard, https://tools.ietf.org/html/rfc6901 , which you should reference. 

A helpful courtesy (need not be part of this pull request) would be to supply a utility that can convert this boneheaded format into something which can be supplied to the ChangeApplier API. Most helpfully, something which will parse both schemaPath and dataPath into an array of path segments which will then make escaping issues moot. A future version of this library could accompany these by schemaSegs and dataSegs in these records holding these parsed values.

the-t-in-rtf: I have updated the docs as suggested.  I also created a ticket for [the follow-up work to better support working with both notations](https://issues.gpii.net/browse/GPII-2152).  IMO this would be a good one to work on the next time the project is "under the knife" (such as when we get around to [future-proofing the internationalisation of error messages](https://issues.gpii.net/browse/GPII-2113)).

the-t-in-rtf: @amb26, just to summarize, even though I don't see any outstanding work to be done here, this should not be merged until we finish with its upstream dependencies.

the-t-in-rtf: As a routine step, I updated all dependencies.  There was some small work to update this to work with webdriver 3.x, which is now complete.  I also moved to the standard linting configuration.

the-t-in-rtf:
When we get around to this, we should review all-tests.js together.  All of the tests required from all-tests.js pass individually, but fail when run together.  I suspect a similar problem in gpii-express-user as well.

the-t-in-rtf:
@amb26, once the gpii-pouchdb pull is merged, it would be good to revisit this one.  It's one of the very oldest, and seemed to be trending towards completion when last we left it.  I will start working on the test stability in advance of that.

the-t-in-rtf:
@amb26, after painstaking troubleshooting, I have confirmed that the test failures only occur when running the tests using `npm test`, and only with `npm` version 3.10.10.  The tests definitely pass when using a 4.x version of `npm`, and pass with `yarn` as well.  As I have no idea how to localize the problem beyond that, I hope it's acceptable to document the incompatibility and let time take care of the rest.  The tests pass in Vagrant, which is configured to use `yarn`.

amb26:
I get the following failure when running the tests by any means:

```
$ yarn test
yarn test v0.19.1
$ node node_modules/.bin/istanbul cover tests/all-tests.js
E:\source\gits\gpii-json-schema\node_modules\.bin\istanbul:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^
SyntaxError: missing ) after argument list
    at Object.exports.runInThisContext (vm.js:76:16)
    at Module._compile (module.js:542:28)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
error Command failed with exit code 1.
```

the-t-in-rtf:
OK, I fixed the problem that was preventing the launch of istanbul, but that only cleared the way to hit the next problem.

I'm currently encountering "address in use" errors not seen on linux or OS X, but only when running the browser tests (they use the same harness as the node tests).  I will continue investigating and comment here when I have more information.

the-t-in-rtf:
The "address in use" errors were a red herring (although they did encourage me to do a bit of review and refactoring of the test environments, which isn't a bad thing).  It turns out that my problem was that chromedriver wasn't in the path on my windows VM.  @amb26, as far as I can see, this is ready for review again.

amb26:
I've installed chromedriver using npm install -g chromedriver, and verified that it is on the PATH by running "chromedriver" from the command line. Unfortunately for some reason it is still not being picked up by the tests. I get the following error:

```
13:16:57.354:  jq: Test concluded - Module "Testing 'evolved' error messages..." Test name "Testing preserving a message in an underlying schema...": 1/1 passed
 - PASS
13:16:57.667:  jq: FAIL: Module "chrome: Testing client-side validation..." Test name "chrome: Validate the simplest valid 'base' record...." - Message: Died on
 test #1     at Object.asyncTest (E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\lib\qunit\js\qunit.js:405:9)
    at E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:651:29
    at Object.fluid.each (E:\Source\gits\gpii-json-schema\node_modules\infusion\src\framework\core\js\Fluid.js:513:17)
    at Object.fluid.test.processTestCase (E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:625:15)
    at E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:674:24
    at Object.fluid.each (E:\Source\gits\gpii-json-schema\node_modules\infusion\src\framework\core\js\Fluid.js:513:17)
    at Object.fluid.test.processTestCaseHolder (E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils. .... [output suppre
ssed at 1024 chars - for more output, increase fluid.logObjectRenderChars]
13:16:57.671:  jq: Source: Error: The ChromeDriver could not be found on the current PATH. Please download the latest version of the ChromeDriver from http://ch
romedriver.storage.googleapis.com/index.html and ensure it can be found on your PATH.
    at Error (native)
    at ServiceBuilder (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\chrome.js:191:13)
    at getDefaultService (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\chrome.js:280:22)
    at Function.createSession (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\chrome.js:700:44)
    at createDriver (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\index.js:167:33)
    at Builder.build (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\index.js:639:16)
    at gpii.webdriver.init (E:\Source\gits\gpii-json-schema\node_modules\gpii-webdriver\src\js\webdriver.js:53:35)
    at togo (E:\Source\gits\gpii-json-schema\node_modules\infusion\src\framework\core\js\FluidIoC.js:1794:33)
    at Object.fire  .... [output suppressed at 1024 chars - for more output, increase fluid.logObjectRenderChars]
13:16:57.675:  Express server listening on port 6985
13:16:57.675:  Express started...
13:16:57.783:  jq: Test concluded - Module "chrome: Testing client-side validation..." Test name "chrome: Validate the simplest valid 'base' record....": 0/1 pa
ssed - FAIL
13:16:58.104:  jq: FAIL: Module "chrome: Testing client-side validation..." Test name "chrome: Validate the simplest valid 'derived' record...." - Message: Died
 on test #1     at Object.asyncTest (E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\lib\qunit\js\qunit.js:405:9)
    at E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:651:29
    at Object.fluid.each (E:\Source\gits\gpii-json-schema\node_modules\infusion\src\framework\core\js\Fluid.js:513:17)
    at Object.fluid.test.processTestCase (E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:625:15)
    at E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUtils.js:674:24
    at Object.fluid.each (E:\Source\gits\gpii-json-schema\node_modules\infusion\src\framework\core\js\Fluid.js:513:17)
    at Object.fluid.test.processTestCaseHolder (E:\Source\gits\gpii-json-schema\node_modules\infusion\tests\test-core\utils\js\IoCTestUti .... [output suppresse
d at 1024 chars - for more output, increase fluid.logObjectRenderChars]
13:16:58.107:  jq: Source: Error: The ChromeDriver could not be found on the current PATH. Please download the latest version of the ChromeDriver from http://ch
romedriver.storage.googleapis.com/index.html and ensure it can be found on your PATH.
    at Error (native)
    at ServiceBuilder (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\chrome.js:191:13)
    at getDefaultService (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\chrome.js:280:22)
    at Function.createSession (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\chrome.js:700:44)
    at createDriver (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\index.js:167:33)
    at Builder.build (E:\Source\gits\gpii-json-schema\node_modules\selenium-webdriver\index.js:639:16)
    at gpii.webdriver.init (E:\Source\gits\gpii-json-schema\node_modules\gpii-webdriver\src\js\webdriver.js:53:35)
    at togo (E:\Source\gits\gpii-json-schema\node_modules\infusion\src\framework\core\js\FluidIoC.js:1794:33)
    at Object.fire  .... [output suppressed at 1024 chars - for more output, increase fluid.logObjectRenderChars]
13:16:58.111:  FATAL ERROR: Uncaught exception: listen EADDRINUSE :::6985
```

amb26:
Looks like selenium-webdriver's detection of chromedriver is bullshit - it looks explicitly for "chromedriver.exe" in the path - https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/chrome.js#L189 - whereas npm install -g chromedriver only places "chromedriver" and "chromedriver.cmd" on the PATH

amb26:
I've filed https://github.com/giggio/node-chromedriver/issues/90 against node-chromedriver. I suggest we just add a brief note in the README under the "chromedriver" heading that this npm module is not to be used and that the user must manually follow the install instructions at https://sites.google.com/a/chromium.org/chromedriver/getting-started to download this to their PATH

the-t-in-rtf:
I have updated the README with caveats about the `chromedriver` install.  Thankfully this is no longer needed on the Vagrant linux box, which now finally has its own package.

the-t-in-rtf:
I also removed the dev dependency on the chromedriver npm package.

PR #7, opened 2017-09-29
GPII-2615: Improve "schema link" support.

the-t-in-rtf:
See [GPII-2615](http://issues.gpii.net/browse/GPII-2615) for details.


the-t-in-rtf:
@amb26, at least on my machine I can reproduce the error catching failure by running `npm test`.  The failing test file runs just fine individually, that command is:

`node tests/js/server/parser-update-failure-tests.js`

the-t-in-rtf:
The `npm test` command also uses instrumented code to prepare code coverage reports. I just confirmed I see the same error running `node tests/all-tests.js` without using Istanbul.

amb26:
Hi - this failure is caused by a conflicting definition within gpii-express' test libraries at https://github.com/GPII/gpii-express/blob/master/tests/js/lib/globalFailureHandler.js - as it seems, this component constructs and makes a global instance unconditionally as soon as the gpii-express test library is included. The problem is then that your definition on line 74 of parser-update-failure-tests.js 

                       event:    "{globalErrorHandler}.events.onError",

binds onto this global object and awaits its firing rather than the one on line 53. So the error is not that your global error handler is not firing in preference to the fatal error handler, but that you are waiting for the firing of the wrong event. This could have been verified with a logging statement at line 53.

You have a few possibilities - 
i) rewrite your gpii-json-schema test so that it uses the definition from gpii-express which looks like was intended to be used as a shared utility
ii) write a more specific selector at line 74 {gpii.tests.schema.globalErrorHandler}.events.onError which ensures to match your own handler rather than gpii-express'
iii) rewrite the utility in gpii-express so that it does not load unconditionally but only on receipt of a directive like gpii.test.express.loadGlobalErrorHandler();
iv) others

the-t-in-rtf:
Thanks very much for figuring that out.  Although option 2 seems the laziest and easiest, options 1 and 3 combined smell better.  For my sins I will search my collected works and confirm whether there are other clones lurking around waiting to hit the same problem and update those as well.

the-t-in-rtf:
@amb26, this pull now uses the updated global failure handler from gpii-express.  I'd like to merge the [related express pull](https://github.com/GPII/gpii-express/pull/35) first so I can use a released version in this pull.

docs/schemaLinks.md
```diff
@@ -0,0 +1,150 @@
...
+returns a set of records when search terms are entered, but which returns an error message when the search terms are 
+invalid or a server error occurs.
+
+## `gpii.schema.handlerWithSchemaHeaders`
+
+An extension of [`gpii.express.handler`](https://github.com/GPII/gpii-express/blob/master/docs/handler.md) that adds
+the HTTP headers when the `sendResponse` or `sendError` invoker is called.  This grade only adds headers, it does not
+decide what success or failure looks like.  It is designed to be overlayed with any existing `gpii.express.handler`
+grade that relies on the standard `sendResponse` and `sendError` functions.  If you are already customising these in
+your handler, this grade will likely not work as expected for you.
```
amb26: This sounds a bit discouraging to the reader. I imagine that there is every chance that the grade works, for example, unless it i) fiddles with the same headers, or ii) fully commits the response before they can be modified. We should rework this paragraph to be clearer about what the conditions really are, and also assist the reader by reworking this  implementation so that rather than functioning by overriding some invokers, instead it consists of some namespaced event handlers which the user can have some fine control over when they act.

src/js/server/schemaLinkHandler.js
```diff
@@ -0,0 +1,47 @@
...
+
+require("gpii-express");
+require("../common/hasRequiredOptions");
+require("./schemaUrlHolder");
+require("./lib/schemaLinkHeaders");
+
+fluid.registerNamespace("gpii.schema.schemaLink.handler");
+
+gpii.schema.schemaLink.handler.sendResponse = function (that, response, statusCode, body) {
+    gpii.schema.schemaLinks.addHeaders(response, that.options.schemaUrls.success);
```
amb26: This is an absolutely shining candidate for being implemented by an event-driven workflow with namespaced listeners, rather than working by "override and copy" - see above comments

tests/js/server/parser-update-failure-tests.js
```diff
@@ -14,43 +14,27 @@ kettle.loadTestingSupport();
...
-    events: {
-        onError: null
-    }
-});
-
-var globalErrorHandler = gpii.tests.schema.globalErrorHandler();
-
-gpii.tests.schema.notifyGlobalError = function () {
-    globalErrorHandler.events.onError.fire(fluid.makeArray(arguments));
+// TODO: Wire this into the listener
```
amb26: Has it not been wired in?

tests/js/server/parser-update-failure-tests.js
```diff
@@ -14,43 +14,27 @@ kettle.loadTestingSupport();
...
+gpii.tests.schema.parser.failure.confirmErrorFired = function (errors) {
+    var matchingError = fluid.find(errors, function (error) {
+        if (error.message && error.message.indexOf("ENOENT") !== -1) {
+            return true;
+        }
+    });
+    jqUnit.assertTrue("There should be an error that matches what we expecte...", matchingError);
```
amb26: The speaker has been cut off before finishing his sentiment

tests/js/server/index.js
```diff
@@ -1,11 +1,13 @@
 "use strict";
 /* eslint-env node */
-require("./parser-update-failure-tests");
+require("./hasRequiredOptions-tests");
 require("./middleware-contentAware-tests");
 require("./middleware-requestAware-tests");
 require("./parser-tests");
+// TODO: Work with Antranig to figure out why the parser errors are not being trapped correctly.
```
amb26: I assume this is now done

the-t-in-rtf: Yes, and removed.

tests/js/server/parser-update-failure-tests.js
the-t-in-rtf: It's a dramatic device to build suspense.  You're supposed to pull the test forward to find a knife sticking out of its back.

Anyway, fixed.

tests/js/server/parser-update-failure-tests.js
the-t-in-rtf: Yes, that was cruft from earlier on in the implementation.  Removed.

src/js/server/schemaLinkHandler.js
the-t-in-rtf: I think I agree, but it seems more in scope for a major (or at least minor) release of gpii-express, as it would require updating everything that uses the current gpii-handler grade.  @amb26, would it be enough for now to write up a ticket for discussion and leave a comment here?

docs/schemaLinks.md
the-t-in-rtf: The latter seems like the right solution longer term, but should be discussed and planned, as it would require quite a few updates through various packages that depend on gpii-express.  I'll work on updating the wording.

the-t-in-rtf: I have updated the wording to indicate the specific things to be aware of in extending the great.  Please review and comment.

src/js/server/schemaLinkHandler.js
the-t-in-rtf: I hope so, because I wrote one up: [GPII-2632](https://issues.gpii.net/browse/GPII-2632).

amb26: Yup, that is great

the-t-in-rtf:
Thanks.  I cut the 1.0.3 release and updated the forward-facing version in master.

PR #1, opened 2015-08-20
GPII-1176:  Initial implementation for review...

the-t-in-rtf:
See [GPII-1176](https://issues.gpii.net/browse/GPII-1176) for background, and check out the README file for details on what the package is meant to do.

PR #11, opened 2018-06-06
GPII-3083: Major refactor to add support for "schema-validated components".

the-t-in-rtf:
See [GPII-3083](https://issues.gpii.net/browse/GPII-3083) for details.


the-t-in-rtf:
@amb26, the work in this pull to add support for schema-validated components (and model components) depends on the new workflow functionality introduced in  FLUID-6148.  Previously, I had to use one of the dev releases that included that work for the tests to pass at all.  Now that that's been merged, I wanted to try updating to pick up the timing data fix in QUnit (and just to keep up to date).

I discovered that the schema-validated component work only works with `3.0.0-dev.20181010T171529Z.aeaa6e324.FLUID-6148` or an earlier release from the same branch and not with any of the mainstream dev releases post-merge of FLUID-6148.  I am investigating now, but wanted to ask if there were any significant changes to the workflow functionality that I should be aware of.

the-t-in-rtf:
If I update even to 3.0.0-dev.20181017T162149Z.787f7d5e5 (two days later than the merge), my startup checks are no longer hit at all.

the-t-in-rtf:
OK, as this one is thorny, here are the steps to reproduce:

1. Set a breakpoint here: https://github.com/the-t-in-rtf/gpii-json-schema/blob/GPII-3083/src/js/common/schemaValidatedComponent.js#L20
2. Install the current versions of everything and verify that breakpoint is hit when running `tests/js/common/schema-validated-component-tests.js`.
3. Update infusion to either a later dev release of the FLUID-6148 work, or any mainstream dev release created after October 16th, 2018.
4. Note that the breakpoint is no longer hit at all.

the-t-in-rtf:
In later testing I noticed that even later dev releases of FLUID-6148 don't work.

amb26:
I tested this with the most recent release made from the FLUID-6148 dev branch which is 3.0.0-dev.20181010T171529Z.aeaa6e324.FLUID-6148 and it worked fine[*]. Note that the "workflow" facility is not supported in any version of Infusion released from trunk until the FLUID-6148 branch is merged in.

[*] at least the node-based tests ran. The browser-based tests failed with a "component does not match its own schema" error which I found hard to debug since the browser tests can't be run from the filesystem and default to minified code via testem. Note that there is a further failure on IE11 which you may not care about re "isInteger" - if so, please add excludes to leave this browser out of testem runs

the-t-in-rtf:
@amb26, the browser test failures were genuine, the "shadow" record's components aren't yet merged and can't be validated using the old rules.  I have softened up that check for now and we can discuss how/whether to firm that up again if needed.  In testing this, I reviewed running the browser tests outside of Testem and added instructions to the README.

I also fixed some lingering errors that weren't handled properly and were spitting out raw AJV messages rather than i18n message keys.  This should be ready for further review.

the-t-in-rtf:
Thanks for verifying with a newer dev release of potentia-ii, I did also update to that.

src/js/common/schemaValidatedComponent.js
```diff
@@ -0,0 +1,249 @@
...
+                         */
+                        // TODO: Talk with Antranig about how to model workflows correctly.
+                        "workflows": {
+                            "type": "object"
+                        }
+                    }
+                }
+            }
+        },
+        workflows: {
```
amb26: We should factor this grade at this point so that we can somewhat gracefully support pre-6148 versions of Infusion, probably by positioning validation as a createOnEvent with priority: "first:authoring" listener. This may not be terribly satisfactory, but we should at least do something to actively support old Infusions since so far only gpii-universal and gpii-windows have been validated against the new branch and we may need to deploy this work before we can check out and fix gpii-app. At the very least, we should have a "sniff" in the tests which just skips the validated component checks if we find we're on an old Infusion. The best way of performing the sniff is probably to check whether fluid.registerPotentia is defined.

the-t-in-rtf: I'm thinking of handling this with context awareness and "mix in" grades for pre-6148 and post.  These would be clearly labeled as non-API and not documented outside of the code itself.  I could use the same "check" in both the tests and the components.

The one thing I could use help with is figuring out how to test both pre-6148 and post.  I guess I could make a fake sub-package with a pre-6148 version of infusion and run the package tests from a wrapper file that requires the pre-6148 version there?  Or would that fall afoul of the deduping?

the-t-in-rtf:
This pull now includes the work to make the schema validation middleware work with kettle.  I'll be cutting a dev release of this shortly to exercise with my validation pull against universal.

the-t-in-rtf:
On Friday, @sgithens pointed out that there were still instances of older schema syntax, like `min`, `max`, and `divisibleBy`.  I have updated the underlying GSS schema to disallow additional properties, and cut a new dev release:

2.0.0-dev.20190408T072733Z.1dc9549.GPII-3083

I will update my SR validation pull and fix any lingering invalid schemas shortly.

docs/gss.md
```diff
@@ -0,0 +1,305 @@
...
+```json
+
+{
+    "isValid": false,
+    "errors": [
+        {
+            "dataPath": ["requiredField"],
+            "schemaPath": ["properties", "requiredField", "required"],
+            "rule": { "required": true },
+            "message": "validation-required-field-missing"
```
amb26: This style of message key could be adjusted so it is consistent with the kinds of message keys shown in the previous section describing the "errors" schema field

docs/gss.md
```diff
@@ -0,0 +1,305 @@
...
+var messages = {
+    "validation-invalid-field-type": "The value supplied should be a(n) %rule.type."
+};
+
+var message = fluid.stringTemplate(messages[error.message], error); // The value supplied should be a(n) boolean.
+```
+
+## Schema-validated components
+
+A schema validated component is one whose structure is required to match its "schema" options, which is defined using
```
amb26: Should refer to the dedicated schemaValidatedComponents.md docs here?

docs/schemaValidatedComponent.md
```diff
@@ -0,0 +1,159 @@
+# `gpii.schema.component`
+
+A "schema validated component" is one whose entire structure (including all options) is validated as part of its
```
amb26: Hard to say that it is the "entire" structure - since the only bits of it that will be validated are the bits a schema has been written for!

docs/gss.md
```diff
@@ -0,0 +1,305 @@
...
+
+Someone using or extending your grade must ensure that `options.requiredOption` exists and meets the specified criteria.
+Otherwise, the component will not instantiate, and an error like the following will be thrown:
+
+```text
+Component does not match its own schema, aborting component creation:
+    options -> requiredOption:  This value is required.
+```
+
+## Schema-validated model components
```
amb26: Should add a stub docs plus ref to the dedicated docs here

src/js/common/schemaValidatedComponent.js
```diff
@@ -0,0 +1,310 @@
...
+            "onCreate.validate": {
+                funcName: "gpii.schema.component.validateComponent",
+                args:     ["{that}"]
+            }
+        }
+    });
+
+    fluid.defaults("gpii.schema.component", {
+        gradeNames: ["fluid.component", "fluid.contextAware"],
+        schema: {
```
amb26: Crikey, I really need to implement the "immutable options revolution". Just imagine a separate copy of this hanging around in memory for every instantiated component ........

src/js/common/schemaValidatedComponent.js
```diff
@@ -0,0 +1,310 @@
+/* globals require */
+(function (fluid) {
+    "use strict";
+
+    if (typeof require !== "undefined") {
```
amb26: This is good enough for now but I imagine might anger some people down the line who might eventually want to distribute infusion/this library/others using webpack or some similar scheme. I don't currently know of any such people. Might it not be safer just to sniff if fluid itself is defined as we do elsewhere?

src/js/common/schemaValidatedComponent.js
```diff
@@ -0,0 +1,310 @@
+/* globals require */
```
amb26: I wonder why this passes linting when fluid itself is used as a global on line 310

src/js/common/validator.js
```diff
@@ -1,122 +1,470 @@
...
+     *
+     */
+    gpii.schema.validator.validate = function (toValidate, gssSchema, ajvOptions) {
+        ajvOptions = ajvOptions || gpii.schema.validator.defaultAjvOptions;
 
-        if (!contentValid) {
-            return (gpii.schema.validator.ajv.sanitizeValidationErrors(that, key, that.ajv.errors));
+        // Validate the GSS schema against the metaschema before proceeding.
+        var schemaValidationResults = gpii.schema.validator.validateSchema(gssSchema, ajvOptions);
+        if (fluid.get(schemaValidationResults, "isError")) {
```
amb26: I imagine a simple schemaValidationResults.isError is good enough since the validator always guarantees to return an object

src/js/common/validator.js
```diff
@@ -1,122 +1,470 @@
...
+
+
+    /**
+     *
+     * Standardise the path to the failing rule within a GSS schema based on a raw AJV validation error.
+     *
+     * NOTE: This function is a non-API function, i.e. one that assists public functions in performing their work, but
+     * which is not guaranteed to remain available.
+     *
+     * @param {Object} ajvError - A single raw validation error as returned by AJV.
```
amb26: Can we have more specific eslint type here to explain the contents of the error structure - using a @typedef - (note to ourselves - someday we will have to fold the capabilities of eslint in describing structures together with those of JSON schema itself)

src/js/common/validator.js
```diff
@@ -1,122 +1,470 @@
...
-        if (!contentValid) {
-            return (gpii.schema.validator.ajv.sanitizeValidationErrors(that, key, that.ajv.errors));
+        // Validate the GSS schema against the metaschema before proceeding.
+        var schemaValidationResults = gpii.schema.validator.validateSchema(gssSchema, ajvOptions);
+        if (fluid.get(schemaValidationResults, "isError")) {
+            return schemaValidationResults;
         }
-    }
+        else {
+            var ajv = new AJV(ajvOptions);
```
amb26: In time we will need to come up with effective means of caching not only these AJV objects but also the resulting compiled schemas since the cost of reconstructing them all, e.g., on every model change, will become prohibitive. I imagine we will make the principal implementation of the validator structure part of component grades and hence supply a place to cache them inside a component, rather than layering our component-directed validation machinery on top of pure functions like this.

src/js/common/validator.js
```diff
@@ -1,122 +1,470 @@
...
+     *  }
+     * }
+     *
+     * When using the "long" notation, you can still fail over to a generic message using the key `""`, as shown above.
+     *
+     * NOTE: This function is a non-API function, i.e. one that assists public functions in performing their work, but
+     * which is not guaranteed to remain available.
+     *
+     * @param {Array<String>} rulePath - An array of EL path segments.
+     * @param {Object} gssSchema - A GSS schema.
```
amb26: Is there anything sensible we can do with a @typedef in order to describe what could go in this object? I imagine in time we could make a "JSON schema to eslint @typedef compiler" that could generate such docs from hints and rules.

src/js/common/validator.js
```diff
@@ -1,122 +1,470 @@
...
+    };
+
+    /**
+     *
+     * Convert a GSS schema to a standard JSON Schema.
+     *
+     * NOTE: This function is a non-API function, i.e. one that assists public functions in performing their work, but
+     * which is not guaranteed to remain available.
+     *
+     * @param {Object} originalFsl - The FSL schema.
```
amb26: Presumably FSL here needs to be renamed GSS

tests/js/common/metaschema-tests.js
```diff
@@ -0,0 +1,100 @@
+/*
+
+    Independently of the validator, confirm that our GSS metaschema is itself valid, and that it can be used with Ajv
+    to validate GSS schemas.
+
+*/
+/* globals Ajv, jqUnit, require */
+/* eslint-env browser */
+var fluid_3_0_0 = fluid_3_0_0 || {};
```
amb26: I think there isn't a need to use the versioned namespace here since this is only of interest to libraries who wish to be versioned along with infusion (which in any case is a scheme we have not bothered to update for many years). The use of plain "fluid" will be fine.

amb26:
Since testem isn't listed as a dependency I got the following error when running the tests:
````
> gpii-json-schema@2.0.0 test:browser E:\source\gits\gpii-json-schema
> node node_modules/testem/testem.js ci --file tests/testem.js

internal/modules/cjs/loader.js:582
    throw err;
    ^

Error: Cannot find module 'E:\source\gits\gpii-json-schema\node_modules\testem\testem.js'
````
Is nyc required any more as a direct dependency?

File: src/js/common/schemaValidatedComponent.js (hunk 8 above)
Comment (the-t-in-rtf) [reply src/js/common/schemaValidatedComponent.js]: It passes because we have [a rule in our ESLint config to indicate that fluid is always a global](https://github.com/fluid-project/eslint-config-fluid/blob/master/.eslintrc.json#L7).

File: src/js/common/schemaValidatedComponent.js (hunk 7 above)
Comment (the-t-in-rtf) [reply src/js/common/schemaValidatedComponent.js]: I updated to check for the presence of fluid instead.

File: src/js/common/validator.js (hunk 11 above)
Comment (the-t-in-rtf) [reply src/js/common/validator.js]: My previous approach was very much along those lines, and I will think through doing something like that again.  In the mean time, I hope it will be sufficient to write up the improvement as a JIRA issue.

File: src/js/common/validator.js (hunk 13 above)
Comment (the-t-in-rtf) [reply src/js/common/validator.js]: I searched for and replaced all variations throughout the project, including this one.

File: tests/js/common/metaschema-tests.js (hunk 14 above)
Comment (the-t-in-rtf) [reply tests/js/common/metaschema-tests.js]: I standardised and cleaned this up throughout.

**Comment by the-t-in-rtf** [comment]
@amb26 the error with the missing testem dependency is luck of the draw depending on where npm stuffs the inherited dependency coming from gpii-testem.  To make testem (and nyc) behave consistently, both must be dev dependencies of this package itself, which I've done now.

File: src/js/common/validator.js (hunk 9 above)
Comment (the-t-in-rtf) [reply src/js/common/validator.js]: Done.

File: docs/schemaValidatedComponent.md (hunk 4 above)
Comment (the-t-in-rtf) [reply docs/schemaValidatedComponent.md]: I ended up with better wording in the summary in the gss.md file and reused key bits to improve this.

File: docs/gss.md (hunk 3 above)
Comment (the-t-in-rtf) [reply docs/gss.md]: Very good catch, seems like  broke it out and then didn't clean up.  Fixed now.

File: docs/gss.md (hunk 5 above)
Comment (the-t-in-rtf) [reply docs/gss.md]: Done.

File: docs/gss.md (hunk 2 above)
Comment (the-t-in-rtf) [reply docs/gss.md]: I (lazily) let this package use non-namespaced keys for its own messages, so that is the actual full key at the moment.  Seems like I should just namespace that with gpii.schema.messages?

File: src/js/common/validator.js (hunk 10 above)
Comment (the-t-in-rtf) [reply src/js/common/validator.js]: This should be a good warmup for sketching out a typedef for the GSS metaschema itself, so I'll give it a go.

File: src/js/common/validator.js (hunk 12 above)
Comment (the-t-in-rtf) [reply src/js/common/validator.js]: @amb26, I have doubts that we can do represent the full range of what's possible with GSS and/or JSON schemas in JSDoc typedefs.  Up front, some things do seem doable slash easy:

1.  We can represent deep @property entries using a very familiar dot notation.
2. Optionality is trivial to represent, i.e. square braces.
3. Type mapping of JSON schema primitives should be straightforward.
4. Recursion (which we need for things like the "schema" field in each setting entry) seems to be supported, although some things that consume JSDocs [don't always handle it well](https://github.com/Microsoft/TypeScript/issues/15597).

I'm a bit more concerned about:

1. Usages of `oneOf`, i.e. "must match only one of these schemas".  Anything more complex than "{String|Array<String>}" is likely to be difficult to express unless we want to keep track of and generate typedefs for each allowed sub-type.
2. Usages of `allOf`, i.e. "must match all of these schemas".  This will involve merging material ourselves to detect the full range of rules and sub-properties.
3. Usages of `anyOf`, i.e. "must match one or more of these schemas".  If we have 3 possibilities, we might need to express 6 sub-typedefs before we can safely "or" them together in JSDoc syntax.

Just to organise my thinking and give me better specific examples, I plan to sketch out a typedef for GSS.  If it proves too difficult, a link to the documentation seems appropriate.

File: src/js/common/validator.js (hunk 10 above)
Comment (the-t-in-rtf) [reply src/js/common/validator.js]: OK, I've committed a typedef for this.

File: src/js/common/validator.js (hunk 12 above)
Comment (the-t-in-rtf) [reply src/js/common/validator.js]: @amb26, I have committed a first pass at this.  If you have even a little time to chat tomorrow we could discuss further defanging some of the complex bits of the JSON Schema language, writing everything out got me thinking.

**Comment by the-t-in-rtf** [comment]
@amb26, as this is a long running pull with a lot of comments, I reviewed to see what we still have left to do/discuss:

1. [I asked you about namespacing the default message keys](https://github.com/GPII/gpii-json-schema/pull/11#discussion_r278484448).
2. [I asked you whether documenting the need for new validation component was adequate](https://github.com/GPII/gpii-json-schema/pull/11#discussion_r278064125).
3. [Your feedback on the new typedefs is needed](https://github.com/GPII/gpii-json-schema/pull/11#discussion_r278562670).
4. [We should have a brief discussion or thread about what if anything we should simplify in the underlying JSON Schema language](https://github.com/GPII/gpii-json-schema/pull/11#discussion_r278562670).  I will comment there with ideas.

File: docs/gss.md (hunk 2 above)
Comment (amb26) [reply docs/gss.md]: That does sound reasonable - although my primary original question was about consistency in the docs itself - the earlier message bundle uses package qualification plus camelCasing e.g. my.package.messages.emailHint - whereas this section using hyphens and no packages. As well as the actual namespacing fix, we should pick a single style for examples and impl

Comment (amb26) [reply docs/gss.md]: However, we should revisit again our earlier discussions with the Bulgarians about I18N formatting and our "various forms of LOADERS" referenced in the following pad: https://pad.gpii.net/p/i18n-meetings-6294n3p - For example the mere fact that you have some stuff in a globally namespaced constant should count for something, and then again we also want to make life easier for users of particularly frequently-referenced messages.

File: src/js/common/validator.js [hunk 15]
```diff
@@ -1,122 +1,540 @@
...
         }
+    };
+
+    /**
+     *
+     * Validate a "GPII Schema System" schema.
+     *
+     * @param {GssSchema} gssSchema - A GSS schema definition.
+     * @param {Object} ajvOptions - Optional arguments to pass to the underlying AJV validator.
+     * @return {Object} - An object that describes the results of validation.  The `isValid` property will be `true` if the data is valid, or `false` otherwise.  The `isError` property will be set to `true` if there are validation errors.
```
amb26: Presumably this should use the newly created @typdef

src/js/common/validator.js
```diff
@@ -1,122 +1,540 @@
...
-            that.ajv.addSchema(schemaContent, schemaKey);
+        // Validate the GSS schema against the metaschema before proceeding.
+        var gssSchemaValid = ajv.validateSchema(gssSchema);
+        if (gssSchemaValid) {
+            return {};
         }
-        catch (e) {
-            fluid.fail("There was an error loading one of your JSON Schemas:", e);
+        else {
+            return { isError: true, message: "Invalid GSS Schema:\n" + JSON.stringify(ajv.errors, null, 2) };
```
amb26: Is it necessarily helpful to flatten all the errors here? The upstream consumer may be interested in parsing them. Perhaps a good compromise is attaching them in a structured form in an "errors" member of the return as well as sticking them on the message

src/js/common/validator.js
amb26: Sorry, ignore earlier comment - I see that this is an entirely different structured return, which nonetheless should have a @typedef of its own

src/js/common/validator.js
amb26: I think the first pass is more than good enough to be going along with, given the purpose of these @typedefs for a very long time will just be to aid orientation and reader comprehension. Just remember to actually change the @param type here to @GssSchema

src/js/common/validator.js
```diff
@@ -1,122 +1,540 @@
...
+     *
+     * Note that as in the examples above, the GSS metaschema defines "error hints" that result in message keys
+     * being returned.
+     *
+     * NOTE: This function is a non-API function, i.e. one that assists public functions in performing their work, but
+     * which is not guaranteed to remain available.
+     *
+     * @param {Object} gssSchema - A GSS schema.
+     * @param {ajvErrors|Boolean} ajvErrors - The raw errors returned by AJV, if there are any, or `false` if there are no validation Errors.
+     * @return {Object} - An object detailing the validation results (see above).
```
amb26: @typedef for this validation return payload here using the newly created @ajvError type

src/js/common/validator.js
```diff
@@ -1,122 +1,540 @@
...
+    /**
+     *
+     * A function to translate/localise validation errors.
+     *
+     * If you want to pass a custom message bundle to this function, it should only contain top-level elements, see
+     * ./src/js/validation-errors.js in this package for an example.
+     *
+     * @param {Array<Object>} validationErrors - An array of validation errors, see `gpii.schema.validator.standardiseAjvErrors` for details.
+     * @param {Any} validatedData - The (optional) data that was validated.
+     * @param {Object} messages - An (optional) map of message templates (see above).  Defaults to the message bundle provided by this package.
```
amb26: I think we can help here with some compound type such as {Object.<String, String>} ?

src/js/common/validator.js
```diff
@@ -1,122 +1,540 @@
...
+     * @param {Object} localisationTransform - An optional set of rules that control what information is available when localising validation errors (see above).
+     * @return {Array<Object>} - The validation errors, with all message keys replaced with localised strings.
```
amb26: Array of something more specific here

src/js/common/validator.js
amb26: Absolutely

the-t-in-rtf:
For discussion today if possible, here are quick thoughts about JSON Schema keywords we can probably safely lose:

1. The `if`/`then`/`else` keywords radically increase the complexity of generating UIs and of displaying error messages to the user,  IMO we should not be writing schemas that complex if we can at all help it (and I think we can).
2. The `readOnly` keyword seems like something we might find interesting if we wanted to represent the structure of an object, but seem less useful for representing preference data that by its nature is changing.  We should only use this if we plan to have tools that are aware of and that respect the "read only" flag.
3. The `patternProperties` keyword seems problematic for UI generation, and hard to provide feedback on.  I would argue we should be writing longer schemas with individually documented entries rather than using pattern properties, unless we want to document the structure of things like generated component trees.
4. The `propertyNames` keyword (rules about property keys rather than property values)  is similarly problematic for UI generation, where properties with a single clear name are a lot easier to model and provide feedback on.
5. The `contentEncoding` and `contentMediaType` keywords might be useful for encoding binary objects (fonts, images, etc.) as strings, but if we plan to store URIs pointing to content rather than content, we can safely lose these.

src/js/common/validator.js
the-t-in-rtf: Agreed, and it greatly simplifies the message as well.  Done.

src/js/common/validator.js
the-t-in-rtf: Done.

src/js/common/validator.js
the-t-in-rtf: Great, see [GPII-3899](https://issues.gpii.net/browse/GPII-3899) for my initial writeup.

docs/gss.md
the-t-in-rtf: Yes, we may need to chat through this one again, as my current approach (passing in a bundle with namespaced keys) is not really compatible with using a namespaced key that can be resolved to a template by way of a global variable.  I've committed a take on this that at least standardises the namespacing and casing for further discussion.

src/js/common/validator.js
the-t-in-rtf: Done.

src/js/common/validator.js
the-t-in-rtf: Yup, done.

src/js/common/validator.js
the-t-in-rtf: Yup.

the-t-in-rtf:
@amb26, I hope I've addressed most of the typedef stuff, we just need to chat about the simplification of GSS (see comment above) and about the namespacing and resolution of i18n keys (perhaps a topical meeting would be useful).

the-t-in-rtf:
@amb26 I left out a relevant comment from an earlier draft [of my comments above regarding simplifying the GSS language](https://github.com/GPII/gpii-json-schema/pull/11#issuecomment-488977400).  The `const` keyword specifies a single allowed value, it's like a single-element `enum`.  I think we should deprecate it in favour of  `enum` because then an `enumLabel` is required.

amb26:
I agree with all of the above recent points - that is, eliminating constructs such as if/then/else, patternProperties, propertyNames, contentEncoding, contentMediaType and const from the basic profile we plan to support for GSS

amb26:
"readOnly" is as you say an interesting middle ground - it is an important feature of the Infusion runtime that some properties are mutable, but we already capture this in your existing usage of component validation which validates options (immutable) one time up front at construction time, and model contents (mutable) on every change. So indeed this can also be pushed out of scope for now

src/js/common/validator.js
amb26: Hasn't been updated yet

the-t-in-rtf:
@amb26 and @sgithens, I have written up my thoughts on simplifying the GSS language in [GPII-3900](https://issues.gpii.net/browse/GPII-3900).  The only open topic I can see from the previous conversations is i18n/message handling, I leave it to you to decide if that needs to happen before this is merged.

src/js/common/validator.js
the-t-in-rtf: Sorry, I reviewed all `{Object}` references but missed the `<Object>` ones.  Actually fixed now.

amb26:
Failure in browser test cases under Firefox in Windows:
````
Died on test #1 asyncTest@http://localhost:7357/node_modules/infusion/tests/lib/qunit/js/qunit.js:405:3
fluid.test.processTestCase/<@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:732:29
fluid.each@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:565:17
fluid.test.processTestCase@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:706:9
fluid.test.processTestCaseHolder/<@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:755:13
fluid.each@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:565:17
fluid.test.processTestCaseHolder@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:748:9
fluid.test.testEnvironment.runTests/<@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:804:13
fluid.each@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:565:17
fluid.test.testEnvironment.runTests@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:796:9
togo@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:2269:24
fire@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:1645:31
fluid.concludeComponentInit@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1363:9
fluid.concludePotentiaePhase/<@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1766:21
fluid.concludePotentiaePhase@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1765:66
fluid.commitPotentiaePhase@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1815:16
commitPotentiae@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1840:35
fluid.tryCatch@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:238:24
fluid.commitPotentiae@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1837:9
fluid.initFreeComponent@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:2751:22
creator@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:2070:20
fluid.invokeGlobalFunction@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:1251:20
fluid.test.makeTestRunner/that.next@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:834:17
fluid.test.makeTestRunner/that.append@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:847:17
fluid.test.runTests@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:866:9
@http://localhost:7357/tests/browser-fixtures/js/errorBinder-tests.js:260:5
@http://localhost:7357/tests/browser-fixtures/js/errorBinder-tests.js:3:2
: Handlebars.registerHelper is not a function

Source: | gpii.handlebars.renderer.common.registerHelpers/<@http://localhost:7357/node_modules/gpii-handlebars/src/js/common/renderer.js:32:17 fluid.each@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:569:17 gpii.handlebars.renderer.common.registerHelpers@http://localhost:7357/node_modules/gpii-handlebars/src/js/common/renderer.js:31:13 togo@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:2269:24 fire@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:1645:31 fluid.concludeComponentInit@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1363:9 fluid.concludePotentiaePhase/<@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1766:21 fluid.concludePotentiaePhase@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1765:66 fluid.commitPotentiaePhase@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1815:16 commitPotentiae@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1840:35 fluid.tryCatch@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:238:24 fluid.commitPotentiae@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1837:9 concludeListener@http://localhost:7357/node_modules/infusion/src/framework/core/js/FluidIoC.js:1282:17 fire@http://localhost:7357/node_modules/infusion/src/framework/core/js/Fluid.js:1645:31 fluid.test.makeFuncExpander/</expanded.apply@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:185:24 execute@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:242:17 fluid.test.execExecutor@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:520:9 fluid.test.sequenceExecutor/that.execute@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:658:21 testFunc@http://localhost:7357/node_modules/infusion/tests/test-core/utils/js/IoCTestUtils.js:714:73 run@http://localhost:7357/node_modules/infusion/tests/lib/qunit/js/qunit.js:207:4 run/<@http://localhost:7357/node_modules/infusion/tests/lib/qunit/js/qunit.js:365:5 process@http://localhost:7357/node_modules/infusion/tests/lib/qunit/js/qunit.js:1457:4 start/<@http://localhost:7357/node_modules/infusion/tests/lib/qunit/js/qunit.js:483:5
````

the-t-in-rtf:
No failure in FF@66.0.2 or FF@66.0.4 in OS X, spinning up a Windows 10 VM now.

the-t-in-rtf:
@amb26, the tests also pass in a Windows 10 VM with FF@64.0.2, upgrading to FF@66.0.4 also works.

the-t-in-rtf:
The problem was with npm versions and the handlebars dependency.  Depending on where npm decides to stuff handlebars (itself a dependency of a few dependencies), it might not be found when the tests are loaded.  Adding it as a direct (dev) dependency should address the issue.

PR #12, opened 2019-02-12
NOJIRA bumping versions of gpii-express and gpii-handelbars to get gp…

sgithens:
bumping versions of gpii-express and gpii-handelbars to get gpii-express 1.0.15


amb26:
Closed since #21 is sufficiently modern

PR #14, opened 2019-05-16
GPII-3920: Updated handlebars to address handlebars security vulnerability.

the-t-in-rtf:
See [GPII-3920](https://issues.gpii.net/browse/GPII-3920) for details.


the-t-in-rtf:
Per discussions with @amb26, we have agreed that I'd self-merge and cut a release for this change.

PR #16, opened 2019-06-18
GPII-3899: Initial implementation of global validation component.

the-t-in-rtf:
See [GPII-3899](https://issues.gpii.net/browse/GPII-3899) for details.


src/js/common/hashString.js
```diff
@@ -0,0 +1,63 @@
...
+     * Strings longer than 50,000 characters will return a hash that looks like multiple stringified 32-bit integers
+     * joined using a colon, as in:
+     *
+     * `1464911337:-123853895:59841156:-143877442`
+     *
+     * @param {String} originalString - The original string to be hashed.
+     * @return {String} - A hash composed of the hashes for each sub-segment in the original string.
+     *
+     */
+    gpii.schema.hashString = function (originalString) {
```
amb26: Whilst you've worked hard to reduce the probability of a hash collision, I feel that in practice we simply can't afford any - at least, "essentially any". The following table explains that, for example, if we are just considering the space of strings shorter than 50k which just get a single 32-bit hash, after only 100,000 strings the chance of a collision becomes greater than 1 in 2. https://stackoverflow.com/questions/14210298/probability-of-collision-when-using-a-32-bit-hash
In any case, the "correct" data structure to be used (either an open or a closed hashtable https://stackoverflow.com/questions/9124331/meaning-of-open-hashing-and-closed-hashing ) would unavoidably store the actual key of the hash somewhere, in order to detect whether a collision has occurred.
In this case I recommend we simply store the key itself - either the JSON.stringify version of the schema or a lightly compacted version thereof, and thus eliminate all the complexity due to this hashing. We can revisit this in future if it becomes a performance worry - but bear in mind that we can't avoid storing the schema itself in memory, and this will at most double that storage cost.

src/js/common/schemaValidatedModelComponent.js
```diff
@@ -16,17 +16,24 @@ var fluid  = fluid  || {};
...
+        var validationResults = globalValidator.validate(modelValidationComponent.options.modelSchema, modelValidationComponent.model, modelValidationComponent.schemaHash);
         // Flag this change as a result of validation so that we can avoid multiple validation passes per model change.
-        that.applier.change("validationResults", validationResults, "ADD", "validation");
+        modelValidationComponent.applier.change("validationResults", validationResults, "ADD", "validation");
     };
 
     fluid.defaults("gpii.schema.modelComponent", {
         gradeNames: ["gpii.schema.component", "fluid.modelComponent"],
+        members: {
+            schemaHash: false
```
amb26: Why not generate the schema hash in an expander in place?

src/js/common/validator.js
```diff
@@ -148,21 +150,47 @@ var fluid  = fluid  || {};
...
-     * @param {Object} ajvOptions - Optional arguments to pass to the underlying AJV validator.
+     * @param {Any} toValidate - The material to be validated.
+     * @param {String} [schemaHash] - An optional unique hash of the schema (used for cache lookups).
      * @return {gssValidationError} - An object that describes the results of validation.  The `isValid` property will be `true` if the data is valid, or `false` otherwise.  The `isError` property will be set to `true` if there are validation errors.
      *
      */
-    gpii.schema.validator.validate = function (toValidate, gssSchema, ajvOptions) {
+    gpii.schema.validator.validate = function (that, gssSchema, toValidate, schemaHash) {
+        if (!schemaHash) {
+            schemaHash = gpii.schema.hashString(gpii.schema.stringify(gssSchema));
```
amb26: See above

src/js/common/validator.js
```diff
@@ -572,4 +596,24 @@ var fluid  = fluid  || {};
...
         return requiredChildren;
     };
+
+    // A global component that validates material using the above static functions.  Handles the expensive initial
+    // compilation of the schema.  For performance reasons, this is not itself a schema-validated component.
+    fluid.defaults("gpii.schema.validator", {
+        gradeNames: ["fluid.component", "fluid.resolveRootSingle"],
+        singleRootType: "gpii.schema.validator",
+        members: {
+            validatorsByHash: {
```
amb26: Provide an operation to evict validators on request otherwise the user has no option but for them to grow without limit!

src/js/common/schemaValidatedModelComponent.js
the-t-in-rtf: My goal was to make it easier to correlate a cached schema with its original component.  If we don't want to use more readable hashes,  I'm not sure there's much value in storing the stringified schema in the component and passing it around.

src/js/common/validator.js
the-t-in-rtf: Agreed, and done and tested.

the-t-in-rtf:
The tests failed because stringify takes longer on the windows VM.  I can relax or remove that check when I'm working on the hashing.

src/js/common/hashString.js
the-t-in-rtf: I'm happy to get rid of it and review if we find that it would be useful to have a shorter hash in the future.

src/js/common/validator.js
the-t-in-rtf: Yup, gone.

the-t-in-rtf:
The last build error was the same timing error in response to a later commit.  The upcoming build should include the timing change.

the-t-in-rtf:
@amb26, I've addressed the first round of comments.

src/js/common/validator.js
```diff
@@ -572,4 +592,28 @@ var fluid  = fluid  || {};
...
+            }
+        },
+        invokers: {
+            validate: {
+                funcName: "gpii.schema.validator.validate",
+                args:     ["{that}", "{arguments}.0", "{arguments}.1", "{arguments}.2"] // gssSchema, toValidate, schemaHash
+            },
+            clearCache: {
+                funcName: "fluid.set",
```
amb26: Whilst clearing the whole cache is probably a worthwhile operation to support, it is more likely that the user will want the more controllable option of evicting a the cache for a particular schema that they know they will not be using again. The operation should accept the gssSchema which is to be evicted, and call a "delete" on the matching cache slot.

the-t-in-rtf: I went just a little further, as I would like to have a place to sketch more performant alternatives off to the side.  I added an invoker to remove a single schema from the cache, and also one to cache a single schema.  I also restored the ability to pass in a schema hash throughout.  This leaves the option to further optimise in derived grades without requiring us to do the work now.

src/js/common/orderedStringify.js
```diff
@@ -0,0 +1,59 @@
+/* globals require */
+var fluid  = fluid  || {};
+(function (fluid) {
+    // TODO: pull this up into Infusion if we need to use this type of object hashing more broadly.
+    "use strict";
+
+    if (!fluid.identity) {
```
amb26: Funny way of checking for Infusion - probably cleanest to move this up to the outer level where we attempt to allocate the fluid global in the first place. I believe that is the form of pattern we use elsewhere.

src/js/common/validator.js
```diff
@@ -148,21 +149,75 @@ var fluid  = fluid  || {};
...
+     *
+     * Add a single GSS schema to the cache.
+     *
+     * @param {Object} that - The validator component itself.
+     * @param {GssSchema} gssSchema - The original GSS schema.
+     * @param {String} [schemaHash] - An optional precomputed hash of the schema.
+     * @return {Object} - The compiled validator created from the GSS schema.
+     *
+     */
+    gpii.schema.validator.cacheSchema = function (that, gssSchema, schemaHash) {
```
amb26: Nice addition - if, as you seem to be allowing in all of these functions, that the user may have computed the hash already, best to produce a generic function name (e.g. gpii.schema.hashSchema, which will currently just be an alias for gpii.schema.stringify) which they might have used to do so -

src/js/common/orderedStringify.js
the-t-in-rtf: I standardised to the usual check for fluid itself and kept the `fluid.require` pattern for the non-fluid dependencies.

src/js/common/validator.js
the-t-in-rtf: Good idea, I added this in the validator file so that it could take advantage of the custom JSDoc object type.  I also updated the docs.

the-t-in-rtf:
The last build failed because grunt wasn't available for linting, I assume a transient provisioning failure.  Ok to test.

the-t-in-rtf:
Second failure, I dug into it, looks like the VM had trouble connecting to chocolatey to download the required packages.

```
==> windows10: [NuGet] An error occurred while loading packages from 'https://chocolatey.org/api/v2/': Unable to connect to the remote server
==> windows10: chromedriver not installed. The package was not found with the source(s) listed.
==> windows10:  Source(s): 'https://chocolatey.org/api/v2/'
```

The process aborted on that install, and missed the step to install grunt as well, which is why the linting checks failed.

the-t-in-rtf:
I see the same problem in [recent windows builds in CI](https://ci.gpii.net/job/windows-create-vm/719/console):

```
python2 not installed. The package was not found with the source(s) listed.
 Source(s): 'https://chocolatey.org/api/v2/'
```

the-t-in-rtf:
Ok to test again.

the-t-in-rtf:
Seems like we're past the networking issues that killed the previous two builds.  @amb26, this should be ready for further review.

the-t-in-rtf:
Thanks, @amb26.  I cut the release and updated the forward-facing version in master.  I also updated the companion pull against universal (will comment there shortly).

PR #18, opened 2019-07-04
GPII-4009: Updated validation middleware schema to cache its schema (and hash) on startup.

the-t-in-rtf:
See [GPII-4009](https://issues.gpii.net/browse/GPII-4009) for details.


src/js/server/kettleValidation.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+});
+
+fluid.defaults("gpii.schema.kettle.request.http", {
+    gradeNames: ["kettle.request.http"],
+    rulesKey: "default",
+    schemaKey: "default",
+    invokers: {
+        handleRequest: {
```
amb26: I don't think we want to intrude on the native request handling workflow of the target component in this way. The original concept was for the validator to be packaged as a piece of statically available middleware in the "eternal".

src/js/server/kettleValidation.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+    },
+    invokers: {
+        validateRequest: {
+            funcName: "gpii.schema.kettle.validator.validateRequest",
+            args: ["{that}", "{gpii.schema.validator}", "{arguments}.0"] // kettleValidator, globalValidator, request
+        }
+    },
+    listeners: {
+        "onCreate.cacheSchemas": {
```
amb26: It will make life easier if, rather than one validator per app, we instead have one per request type. Then we can either have a middleware component as a single static child, or depending on how messy it is, have the validator be the middleware itself.

src/js/server/kettleValidation.js
```diff
@@ -0,0 +1,107 @@
...
+fluid.defaults("gpii.schema.kettle.app", {
+    gradeNames: ["kettle.app"],
+    components: {
+        validator: {
+            type: "gpii.schema.kettle.validator"
+        }
+    }
+});
+
+fluid.defaults("gpii.schema.kettle.request.http", {
```
amb26: I think, as agreed, for ultimate performance we need to entirely axe the schema-specialised request grade. Requests will opt in to validation by making a reference to statically available middleware hosted in the "eternal".

src/js/server/kettleValidation.js
amb26: You could, of course, carry on hosting all validators in a single static component in this way, but this then raises the problem of how you arrange for all the middleware components to be constructed sufficiently early - this could be done but would probably involve something messy such as a synthesised raw dynamic grade definition.

src/js/server/kettleValidation.js
the-t-in-rtf: So, I have been reading up further and looking at the docs and tests in kettle.  If I understand you correctly, we would stuff a bunch of usable-as-middleware instances, one per type of data at the app level, and then refer to them as part of the `requestMiddleware stack` within a kettle.request.http grade.  Similar to [what's done here](https://github.com/fluid-project/kettle/blob/670396acbf4be31be009b2b2dee48373134ea94d/tests/shared/SessionTestDefs.js#L64) but without hosting the middleware component in the kettle.request grade itself.

amb26: Yes, that's correct

PR #19, opened 2019-07-15
GPII-4022: Cleanup following recent kettle refactor.

the-t-in-rtf:
See [GPII-4022](https://issues.gpii.net/browse/GPII-4022) for details.


the-t-in-rtf:
This PR now also includes the work to add a "bad request" (400) status code to invalid requests handled by the express validation middleware.  See [GPII-4098](https://issues.gpii.net/browse/GPII-4098) for details.

src/js/common/validator.js
```diff
@@ -212,6 +212,18 @@ var fluid  = fluid  || require("infusion");
...
 
+    /**
+     *
+     * Clear all cached validators.
+     *
+     * @param {Object} that - The validator component itself.
+     *
+     */
+    gpii.schema.validator.clearCache = function (that) {
+        delete that.validatorsByHash;
```
amb26: You could omit this line with no effect

src/js/server/kettleValidation.js
```diff
@@ -5,6 +5,17 @@ var gpii  = fluid.registerNamespace("gpii");
 
 fluid.registerNamespace("gpii.schema.kettle.validator");
 
+/**
+ *
+ * Validate a request payload according to a GSS Schema.  Fulfills the contract for a `kettle.middleware` `handle`
+ * invoker.
+ *
+ * @param {Object} kettleValidator - A `gpii.schema.kettle.validator` instance that has a schema and rules about which part of the payload should be validated.
```
amb26: I'm tending to prefer putting at the very least {Component} in these cases but also more helpfully, e.g. {gpii.schema.kettle.validator}

src/js/common/validator.js
```diff
@@ -212,6 +212,18 @@ var fluid  = fluid  || require("infusion");
         delete that.validatorsByHash[schemaHash];
     };
 
+    /**
+     *
+     * Clear all cached validators.
+     *
+     * @param {Object} that - The validator component itself.
```
amb26: {gpii.schema.validator} ?

tests/js/common/schema-holder-tests.js
```diff
@@ -0,0 +1,247 @@
...
+    fluid.registerNamespace("gpii.tests.schema.schemaHolder.transforms");
+    gpii.tests.schema.schemaHolder.transforms.mergeChild = function (transformSpec, transformer) {
+        if (!transformSpec.toMerge) {
+            fluid.fail("mergeChild transform requires a `toMerge` inputPath.", transformSpec);
+        }
+        var toMerge = transformer.expand(transformSpec.toMerge);
+        // TODO: pass both components rather than just their schemas, so that we can use some aspect of the component
+        //  rather than the "title" property of the schema.
+        var outputPath = fluid.model.composePaths(transformer.outputPrefix, toMerge.title);
+        transformer.applier.change(outputPath, toMerge);
```
amb26: This is just too much - we can't have a transform rule which reaches out in this side-effectful way to the transformer's machinery

src/js/common/schemaHolder.js
```diff
@@ -0,0 +1,107 @@
...
+        members: {
+            generatedSchema: false,
+            subComponentSchemaPromises: []
+        },
+        schema: {
+            $schema: "gss-v7-full#",
+            additionalProperties: true
+        },
+        invokers: {
+            getSchema: {
```
amb26: whilst we can currently support events and invokers with the same name, I'd feel more comfortable if you distinguished them somehow

tests/js/common/schema-holder-tests.js
the-t-in-rtf: We should review in today's meeting, this was just a first pass.  I need some way of constructing a map of material from child components based potentially on some property of theirs.    I need an approach like this to generate `supportedSettings`, `capabilitiesTransforms`, and `inverseCapabilitiesTransforms` blocks within a "solution component" (or perhaps a "launch handler component") based on whatever child "setttings components" are found.

src/js/common/schemaHolder.js
the-t-in-rtf: That even was left over from previous work, and has been purged.

src/js/common/validator.js
the-t-in-rtf: Done.

src/js/server/kettleValidation.js
the-t-in-rtf: I went through all JSDocs usages of `{Object}` and updated all of them.  In a few cases I did have the grades in the description, I moved most of those to the type.  I left a single redundant one on this line rather than contort the description to avoid mentioning the grade again.

src/js/common/validator.js
the-t-in-rtf: Done.

src/js/common/schemaHolder.js
the-t-in-rtf: In later discussions I noticed that there were in fact two naming collisions, and that only one had been removed.  I renamed the other.

the-t-in-rtf:
@amb26, I am currently blocked getting the "error binder" to work in the "resource loading" world.  First, here's how to see a practical demo of the problem:

1. Run `node node_modules/testem/testem.js --file tests/testem.js`
2. Open the demo page: `http://localhost:7357/tests/browser-fixtures/errorBinder-demo.html`
3. In the console, take a look at the `errorBinder` component.

That component should have validation errors after startup, but for whatever reason its renderer is never created.  I can see that [the `serverResourceAware` grade](https://github.com/the-t-in-rtf/gpii-handlebars/blob/45989f6ea808caf37d849628202e2499c9d82261/src/js/client/serverResourceAware.js#L54) that brings in the renderer is part of the gradeNames stack, but even if I manually fire the `onResourcesLoaded` event that should create the component, the renderer never shows up.  The resource loading bits from that grade show up, as does the `onRendererAvailable` event.

Any insight you can provide or even ideas about threads to pull on would be helpful.

amb26:
Your renderer is created, but has its own construction cancelled by issuing fluid.fail with the error 11:22:16.811:   ASSERTION FAILED:  Renderer can't find template 'error '. which can be seen in the console on startup

the-t-in-rtf:
For future reference, the error was hidden from me by an odd behaviour in which neither Chrome nor Opera would show me the exception.  This was true both for the exception generated during the initial page load, and for the exception that resulted from triggering the `onRendererAvailable` event and recreating the renderer.

Thankfully, Firefox behaves as expected, which is how I fixed the error, starting from the error @amb26 saw, and fixing the additional legitimate errors in my code.

the-t-in-rtf:
The provisioning steps in the "all tests" job failed with a connection error, likely in talking with chocolatey.org.  The VM didn't get to the point where it had `Grunt` and other `npm` dependencies installed, so that failed as well.  Anyway, OK to test again.

the-t-in-rtf:
Thanks, @amb26.  I cut the release, added the release notes, and updated the forward-facing version in master.

PR #20, opened 2019-11-26
FLUID-6145: Trial dependency update to FLUID-6145 era dependencies

amb26:
The node tests run fine, but unfortunately this jams at the point testem tries to start up


amb26:
Closed since #21 has been merged

PR #21, opened 2020-05-12
GPII-4479: Update to work with newer versions of Infusion.

the-t-in-rtf:
See [GPII-4479](https://issues.gpii.net/browse/GPII-4479) for details.


tests/js/common/lib/check-potentia-grades.js
```diff
@@ -0,0 +1,35 @@
...
+
+(function (fluid) {
+    "use strict";
+    if (!fluid.identity) {
+        fluid = require("infusion");
+    }
+
+    var gpii = fluid.registerNamespace("gpii");
+
+    require("../../../../");
```
amb26, line 13: Generally nicer to write fluid.require("%gpii-json-schema") if you can

the-t-in-rtf: Thanks for catching that, it's fixed.

amb26:
Hang invoking testem - ok to test

the-t-in-rtf:
Error late in the VM creation process, which resulted in a usable node but no grunt.  So the VM creation failed, as did the linting.  The actual tests passed.  Ok to test again.

the-t-in-rtf:
Same provisioning error, will try to reproduce locally.

the-t-in-rtf:
Passes locally using the same steps we use in CI.

the-t-in-rtf:
Ok to test again.

the-t-in-rtf:
@amb26, per my recent goof with gpii-handlebars I went ahead and nudged all dependencies.  As soon as this passes CI, IMO we are good to merge and move on to fixing gpii-express-user.

PR #32, opened 2021-01-12
GH-31: Migrate to `fluid-lint-all` and update all dependencies (resolves #31).

### LINKED ISSUE #31: Update to use `fluid-lint-all`
Also update all dependencies we can safely.

the-t-in-rtf:
This fails currently, but is submitted for reference.  There are test errors with the binder grades that I suspect are a variation on the same issues being tracked on [the fluid-handlebars pull](https://github.com/fluid-project/fluid-handlebars/pull/38).


the-t-in-rtf:
@amb26, this fails with recent versions of Infusion.  It failed the same way before, so hopefully this is not actually a variation on the last bug.  In short, the [`gatedModelWatcher`](https://github.com/fluid-project/fluid-json-schema/blob/main/src/js/client/errorBinder.js#L156) that rerenders the form when there are validation errors does not end up having its model populated, and as such can't listen for model changes.  The easiest way to see the problem is to:

1. Launch the (non-instrumented) test harness using a command like: `node tests/js/launch-test-harness.js`
2. Open the error binder demo: `http://localhost:6194/content/errorBinder-demo.html`
3. Open the console and examine `errorBinder.gatedModelWatcher`.

amb26:
This issue (errorBinder fails to instantiate completely) is fixed in my latest branch at 3.0.0-dev.20210114T215306Z.2b1fe9609.FLUID-6580
I think it is probably time to ditch the FLUID-6145 and FLUID-6148 branches since it will take me too long to maintain them, and go along with the FLUID-6580 branches as "current". These contain some provisional code in "integral bindings" but the changes to the core framework are conservative.

the-t-in-rtf:
Self-merging dependency update per discussion with Antranig.

PR #34, opened 2021-01-19
GH-33: Make model-validation component independent of options-validation component (resolves #33).

### LINKED ISSUE #33: Disentangle "options validation" grade from "model validation" grade.
Currently, the "model validation" grade provided by this package is an extension of the "options validation" grade.   We discovered this because the contract for model listeners used by the "model validation" grade is in fact incorrect and does not support the use of `this` instead of `func` or `funcName`.

Given that the framework's contract is likely to change over time, it would be best if people could use "model validation" without having to also use "options validation".

[See FLUID-6592](https://issues.fluidproject.org/browse/FLUID-6592) for previous thoughts on addressing this.

the-t-in-rtf:
This pull makes the model-validation component work without attempting to validate the component's options.  This should resolve issues @jobara encountered using this package.


the-t-in-rtf:
@jobara, I cut a dev release (`2.1.6-dev.20210119T104920Z.64caad3.GH-33`) for you to try in your work that was affected by the broken contract for model listener definitions.

src/js/common/schemaValidatedComponent.js
```diff
@@ -92,211 +92,7 @@ var fluid  = fluid  || require("infusion");
     fluid.defaults("fluid.schema.component", {
         gradeNames: ["fluid.component", "fluid.contextAware"],
         schema: {
-            "$schema": "fss-v7-full#",
-            "definitions": {
```
amb26, line 96: Shouldn't this material have gone into another grade rather than being axed completely?

the-t-in-rtf: No, none of those definitions or properties are necessary to establish your own contract about what options you expect.  All of the rules there and in the model validated component were trying to ensure that basic component material like listeners, et cetera were defined correctly.

amb26: Well that is still valuable material to have expressed somehow, even if it is provisional - could it be retained and tested in a grade labelled as experimental or so?

jobara:
> @jobara, I cut a dev release (`2.1.6-dev.20210119T104920Z.64caad3.GH-33`) for you to try in your work that was affected by the broken contract for model listener definitions.

I just tried the dev release in my branch and was able to use a `"this"` property in a model listener for the wiring up to `console.log`. The component initialized properly and the I was able to log the validation errors using the `validationResults` model path provided by the `fluid.schema.modelComponent` grade.

src/js/common/schemaValidatedComponent.js
the-t-in-rtf: Yes, I can reconstitute it as an experimental grade with its own tests.

the-t-in-rtf: OK, I made a pass at it, this should be ready for review again.

the-t-in-rtf:
Thanks, @amb26.  @jobara , I cut a full release of `2.1.6` that you can use in your work.

